PR #5, opened 2024-03-30
Replace deprecated HMSET with HSET in examples

randommm:
removing deprecated commands

### DIFF
--- a/topics/indexing.md
+++ b/topics/indexing.md
@@ -83,9 +83,9 @@ with the indexed field, it is possible to store just the ID of the object.
 For example I may have Hashes representing users. Each user is
 represented by a single key, directly accessible by ID:
 
-    HMSET user:1 id 1 username antirez ctime 1444809424 age 38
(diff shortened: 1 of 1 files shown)


hwware:
I do not recommend removing any deprecated commands right now. Maybe in later version, we can consider this.

zuiderkwast (approved):
The change is updating an example to avoid use of the deprecated command in the example.

The change is good, but the description in the pull request is very bad IMO. Other reviewers have understood the description as if the documentation for the deprecated command is deleted.

Thanks for the PR and please use a better title and description next time.

stockholmux:
I agree with @zuiderkwast. 
@randommm Would you be able to update the PR description and comment? 

Also, the DCO bot is not happy. Would you be able to sign off your commits?

randommm:
@zuiderkwast should be okay now

zuiderkwast (approved):
OK, looks good.

From the title, it looks like we're removing the documentation of HMSET completely, but we just remove it from some examples. I'll edit the title again.

PR #16, opened 2024-04-08
Remove Redis from docs that are referenced from conf files

madolson:
Generally replaced Redis->Valkey, but also removed versioning information and rewrote the auth section.

This subset covers the pages mentioned in the valkey.conf file, if we publish this page.

### DIFF
--- a/docs/manual/keyspace-notifications.md
+++ b/docs/manual/keyspace-notifications.md
@@ -53,7 +53,7 @@ just the subset of events we are interested in.
 
 By default keyspace event notifications are disabled because while not
 very sensible the feature uses some CPU power. Notifications are enabled
-using the `notify-keyspace-events` of Valkey.conf or via the **CONFIG SET**.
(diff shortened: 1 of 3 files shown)


docs/manual/keyspace-notifications.md
```diff
@@ -53,7 +53,7 @@ just the subset of events we are interested in.
 
 By default keyspace event notifications are disabled because while not
 very sensible the feature uses some CPU power. Notifications are enabled
-using the `notify-keyspace-events` of redis.conf or via the **CONFIG SET**.
+using the `notify-keyspace-events` of Valkey.conf or via the **CONFIG SET**.
```
stockholmux, line 56: should be `valkey.conf` (lower case) right?

stockholmux (changes requested):
one tiny change.

hwware (approved):

PR #19, opened 2024-04-09
Remove "available since" keyword

hwware:


zuiderkwast (changes requested):
1. I don't like to remove these. They are useful. Why remove them at all? Ok very old info like redis 2.0 can be removed, but in other cases we can say like Since Redis OSS 7.0.

2. The module API docs are generated using a script from the comments in module.c so we shal not edit this page manually at all.

stockholmux:
If it's generated @zuiderkwast should we just remove it entirely from the docs repo? I'd say yes.

zuiderkwast:
> If it's generated @zuiderkwast should we just remove it entirely from the docs repo? I'd say yes.

Yes, in the best of worlds... Currently it's manually generated and updated. When we have an automatic process, we can remove it from the doc repo. But it's somehow nice to have all docs in the doc repo though, if we want to be able to build man pages or something else, without requiring the valkey main repo as a dependency.

There is a comment near the top of the file that says that it's generated: https://github.com/valkey-io/valkey-doc/blob/main/docs/reference/modules/modules-api-ref.md?plain=1#L11

stockholmux:
@zuiderkwast Sounds like we should setup a github action package everything up for man pages? Not today, obviously.

zuiderkwast:
Anyway, the module API is not updated very often. For that reason, I think we should focus on other things instead. We should however run that script once to generate this markdown file sometime soon. I believe rebranding of module.c is complete now.

PR #25, opened 2024-04-10
rebrand [Rr]edis->[Vv]alkey, docs/management/_index.md

daniel-house:
A trivial file!

@madolson @zuiderkwast @stockholmux please review


stockholmux (approved):
LGTM!

Although this is just metadata for the old Redis website, which doesn't make sense in context of the new Valkey website (fields are different). Doesn't hurt to keep it, but I don't think we'll actually use a metadata only file.

PR #27, opened 2024-04-12
rebrand [Rr]edis->[Vv]alkey, docs/management/scaling.md

hwware:
In this PR, we update most "redis" to "valkey", and "Redis" to "Valkey"
But we still keep the redis-rb-cluster

### DIFF
--- a/docs/management/scaling.md
+++ b/docs/management/scaling.md
@@ -995,7 +995,7 @@ The command moves all the keys of a running instance (deleting the keys from
 the source instance) to the specified pre-existing Valkey Cluster. 
 
 {{% alert title="Note" color="info" %}} 
-Starting with Redis 5, if not for backward compatibility, the Redis project no longer uses the word slave. Valkey project strongly recommend using replica instead. Unfortunately in this command the word slave is part of the protocol, so we'll be able to remove such occurrences only when this API will be naturally deprecated.
+Starting with Valkey 5, if not for backward compatibility, the Valkey project no longer uses the word slave. Unfortunately in this command the word slave is part of the protocol, so we'll be able to remove such occurrences only when this API will be naturally deprecated.
 {{% /alert %}} 
 
 ## Learn more
--- a/docs/management/scaling.md
+++ b/docs/management/scaling.md
@@ -1,8 +1,8 @@
 ---
-title: Scale with Redis Cluster
-linkTitle: Scale with Redis Cluster
+title: Scale with Valkey Cluster
+linkTitle: Scale with Valkey Cluster
 weight: 6
-description: Horizontal scaling with Redis Cluster
+description: Horizontal scaling with Valkey Cluster
 aliases: [
     /topics/cluster-tutorial,
     /topics/partitioning,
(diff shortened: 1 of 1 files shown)


docs/management/scaling.md
```diff
@@ -976,34 +976,30 @@ in the same way, so no distinction will be made in the documentation.
...
-the source instance) to the specified pre-existing Redis Cluster. However
-note that if you use a Redis 2.8 instance as source instance the operation
-may be slow since 2.8 does not implement migrate connection caching, so you
-may want to restart your source instance with a Redis 3.x version before
-to perform such operation.
+the source instance) to the specified pre-existing Valkey Cluster. 
 
 {{% alert title="Note" color="info" %}} 
-Starting with Redis 5, if not for backward compatibility, the Redis project no longer uses the word slave. Unfortunately in this command the word slave is part of the protocol, so we'll be able to remove such occurrences only when this API will be naturally deprecated.
+Starting with Redis 5, if not for backward compatibility, the Redis project no longer uses the word slave. Valkey project strongly recommend using replica instead. Unfortunately in this command the word slave is part of the protocol, so we'll be able to remove such occurrences only when this API will be naturally deprecated.
```
zuiderkwast, line 998: For massive replacements like this one, I want it to be only redis -> valkey. Otherwise, such changes will drown in the massive diff.

Anyway, there is already a note and adding "strongly recommend" doesn't make things any better IMO. We can change the terminology and remove these notes.

This page has a lot of other problems too, like the prompt and output of `redis-cli` AKA `valkey-cli` does not match what it actually look like in real life, but we can't fix everything in the same PR.

zuiderkwast (changes requested):
This looks mostly right. Changing Redis to Valkey seems like a good way to do this reasonably fast.

(Btw, I always disliked that they (redis labs) renamed this page to "scaling" when it was historically always called "cluster". It was hard to even find it in the menus when you're looking for the cluster documentation.)

hwware:
> This looks mostly right. Changing Redis to Valkey seems like a good way to do this reasonably fast.
> 
> (Btw, I always disliked that they (redis labs) renamed this page to "scaling" when it was historically always called "cluster". It was hard to even find it in the menus when you're looking for the cluster documentation.)

I can not found cluster information easily in redis website before. Because not everyone know scaling almost is cluster. I suggest rename it properly.

docs/management/scaling.md
hwware: no problem, I will first replace Redis to Valkey, then in another PR, we could update others in this page. Thanks

docs/management/scaling.md
```diff
@@ -976,34 +976,30 @@ in the same way, so no distinction will be made in the documentation.
...
-the source instance) to the specified pre-existing Redis Cluster. However
-note that if you use a Redis 2.8 instance as source instance the operation
-may be slow since 2.8 does not implement migrate connection caching, so you
-may want to restart your source instance with a Redis 3.x version before
-to perform such operation.
+the source instance) to the specified pre-existing Valkey Cluster. 
 
 {{% alert title="Note" color="info" %}} 
-Starting with Redis 5, if not for backward compatibility, the Redis project no longer uses the word slave. Unfortunately in this command the word slave is part of the protocol, so we'll be able to remove such occurrences only when this API will be naturally deprecated.
+Starting with Valkey 5, if not for backward compatibility, the Valkey project no longer uses the word slave. Unfortunately in this command the word slave is part of the protocol, so we'll be able to remove such occurrences only when this API will be naturally deprecated.
```
zuiderkwast, line 998: There is no Valkey 5 (and I think referring to Redis 5 is not relevant here).

```suggestion
If not for backward compatibility, the Valkey project no longer uses the word slave. Unfortunately in this command the word slave is part of the protocol, so we'll be able to remove such occurrences only when this API will be naturally deprecated.
```

zuiderkwast (approved):
OK, I want to merge this and finish the rebranding soon.

Issue #18, opened 2024-04-09
Update references from Redis to Valkey


madolson:
We need help updating the documentation to update all references of Redis to Valkey outside of the command references (which was previously completed in https://github.com/valkey-io/valkey-doc/pull/4). When updating the documentation, watch out for the following:
1. Only update independent references to Redis. Links to "redis.io" can usually be updated to "valkey.io" because the pages are organized in the same way (the contents of this doc repo). Don't update parts of other URLs or references to external websites. Feel free to delete references to Redis inc. products like RedisJSON as appropriate.
2. Generally keep the capitalization.  "Redis" -> "Valkey" and "redis" -> "valkey". This works for almost all usage, for example `redis-cli` becomes `valkey-cli`. Some usages of lowercase "redis" need to be handled specially though:
   * The `redis` object in Lua scripts is replaced by `server`. Examples:
     ```diff
     -EVAL_RO "return redis.call('GET', KEYS[1])" 1 mykey
     +EVAL_RO "return server.call('GET', KEYS[1])" 1 mykey
     ```
     ```diff
     -FUNCTION LOAD "#!lua name=mylib \n redis.register_function('myfunc', function(keys, args) return args[1] end)"
     +FUNCTION LOAD "#!lua name=mylib \n server.register_function('myfunc', function(keys, args) return args[1] end)"
     ```
   * Examples where "redis" is part of the valkey-cli prompt are just wrong. The real prompt is just the host and port.
     ```diff
     -redis 127.0.0.1:6379> scan 0
     +127.0.0.1:6379> scan 0
     ```

4. Skip the names of clients and tools. "StackExchange.Redis" is obviously the name of a client. It's best to skip the whole directories `clients/`, `modules/` and `tools/` when doing the bulk of this work and we'll handle them separately. (We need to investigate, discuss and decide what to do with them.)
5. Explicit references to old version of Redis are handled in either of these ways, depending on the context and usefulness of the information:
   * If it's references to Redis 1-5 (really old), they are simply deleted, but sometimes the sentences need to be rewritten to make sense. Examples:
      * ```diff
        -Starting with Redis 3.2, it is possible to
        +It is possible to
        ```
      * ```diff
        -Firstly, and ever since Redis 2.6.0, the `EVAL` command enables running server-side scripts.
        +Firstly, the `EVAL` command enables running server-side scripts.
        ```        
      * Delete the whole section or note if it's only valid for the really old versions.
        ```diff
[...]

PR #39, opened 2024-04-22
Rebrand commands/*.md

zuiderkwast:
Commands for automatic replacements, one per commit:

```
git grep-replace -P '^redis 127.0.0.1:6379>' '127.0.0.1:6379>'
git grep-replace -F 'https://redis.io/topics/' 'https://valkey.io/topics/'
git grep-replace -P 'redis(?=\.call|\.pcall|\.regis|\.set_repl)' server
git grep-replace -P 'Redis(?= compatibility| [2-7]| version)' 'Redis OSS'
```

All other occurrences of "Redis" were already replaced by "Valkey" in earlier PRs.

Part of #18.


zuiderkwast:
@SoulPancake FYI

commands/blpop.md
```diff
@@ -57,7 +57,7 @@ What happens is that the command performing multiple pushes is executed, and *on
     Client A:   BLPOP foo 0
     Client B:   LPUSH foo a b c
 
-If the above condition happens using a Redis 2.6 server or greater, Client **A** will be served with the `c` element, because after the `LPUSH` command the list contains `c,b,a`, so taking an element from the left means to return `c`.
+If the above condition happens using a Redis OSS 2.6 server or greater, Client **A** will be served with the `c` element, because after the `LPUSH` command the list contains `c,b,a`, so taking an element from the left means to return `c`.
```
PingXie, line 60: "OSS Redis" or "Redis OSS"?
```suggestion
If the above condition happens using a OSS Redis 2.6 server or greater, Client **A** will be served with the `c` element, because after the `LPUSH` command the list contains `c,b,a`, so taking an element from the left means to return `c`.
```

PingXie (approved):
LGTM overall. I am more in favor of "OSS Redis" as opposed to "Redis OSS" though

commands/blpop.md
zuiderkwast: IANAL but I believe we must to use "Redis OSS" according to https://redis.io/legal/trademark-policy/, or maybe even "Redis® OSS".

> Offering an XYZ software, which is an altered, modified, or combined copy of Redis OSS, including Redis as a cloud service or as a container service fully complying with the Redis OSS API. In the product description, you may state that “XYZ software is compatible with the Redis® OSS API” or “Compatible with legacy Redis® OSS.” No other term or description of your software is allowed.

(...)

> Attribution. Any permitted use of the Trademarks, as indicated above, must comply with the following provisions:
>
> - You should always add the R mark (®) to “Redis”, and an asterisk (*) to the first mention on a website as part of a product description;
> - On each page where “Redis®*” is used – add the following legend in a noticeable and readable format: “*Redis is a registered trademark of Redis Ltd. Any rights therein are reserved to Redis Ltd. Any use by <company XYZ> is for referential purposes only and does not indicate any sponsorship, endorsement or affiliation between Redis and <company XYZ>”; and
> - Sections i. and ii. above apply to any appearance of the word “Redis” in: (a) any web page, gated or un-gated; (b) any marketing collateral, white paper, or other promotional material, whether printed or electronic; and (c) any advertisement.

I can't say I understand the extents of this, but possibly it applies only to a product descriptions and open source software are not products, but what do i know...

zuiderkwast: We should remove the references to 2.x though, but I did this PR using automatic replacements, as mentioned. We can go through them and remove them later. If the trademark policy makes it hard, we can just remove all references, but let's find out first.

zuiderkwast: I merged this now, but it's easy to replace all occurrences of "Redis OSS" whenever we know what we have to write.

PingXie: SGTM. 

>I can't say I understand the extents of this, but possibly it applies only to a product descriptions and open source software are not products, but what do i know...

I can go with "Redis OSS" too as long as we are consistent.

> We should remove the references to 2.x though, but I did this PR using automatic replacements, as mentioned.

I noticed that too but agreed with you that we should do one thing at a time.

PR #64, opened 2024-04-25
Change a few remaining occurrences of Redis

### LINKED ISSUE #18: Update references from Redis to Valkey

zuiderkwast:
Fixes #18


stockholmux (approved):
LGTM

PR #67, opened 2024-04-25
updated module docs to reference Valkey

dmitrypol:
I did not just globally search and replace everything but only updated docs that are relevant.  If you see a mistake, please point it out and I will fix it.


zuiderkwast (approved):
Thanks!

Some of these files are generated, but it doesn't matter. If we generate them again, they'll just be overwritten and that's fine.

PR #72, opened 2024-05-01
Update client-side-cacing.md

Virusuki:
If i don't look closely the explanation in client-side-caching, I think the Database term is ambiguous.
Because it can be thought of as the entire database, it would be better to specify it as a valkey.

Also, change channel name the __redis__:invalidate to __valkey__:invalidate.

### DIFF
--- a/topics/client-side-caching.md
+++ b/topics/client-side-caching.md
@@ -10,26 +10,26 @@ aliases:
 
 Client-side caching is a technique used to create high performance services.
 It exploits the memory available on application servers, servers that are
-usually distinct computers compared to the valkey nodes, to store some subset
-of the valkey information directly in the application side.
+usually distinct computers compared to the Valkey nodes, to store some subset
+of the Valkey information directly in the application side.
 
-Normally when data is required, the application servers ask the valkey about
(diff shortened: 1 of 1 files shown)


zuiderkwast (changes requested):
It would be a breaking change to change client-side caching like that. (Obviously you didn't test it. :slightly_smiling_face:) There's a PR about it, but it's not yet decided how we can do it to avoid breaking clients.

Changing the word database in the text is OK, but the name is Valkey, not valkey.

Virusuki:
> It would be a breaking change to change client-side caching like that. (Obviously you didn't test it. 🙂) There's a PR about it, but it's not yet decided how we can do it to avoid breaking clients.
> 
> Changing the word database in the text is OK, but the name is Valkey, not valkey.


I know there are side effects to changing __redis__:invalidate.
I think I'll have to think carefully about it PR [#410](https://github.com/valkey-io/valkey/pull/410).

Additionally, I changed it to "database" -> "Valkey".

zuiderkwast (approved):
Thanks! It like to merge this change first.

PR #74, opened 2024-05-02
Delete guides for Redis official clients

zuiderkwast:
* These clients are Redis official clients, not Valkey
* Guides like these are hard to maintain. Examples like these are normally in the client's own README or their own docs. We'll link to those in the list of clients.
* Even if we my want such guides on the website in the future, I doubt they belong in the doc repo. (They're not a good fit for man pages.)


hwware (approved):

madolson:
> Even if we my want such guides on the website in the future, I doubt they belong in the doc repo. (They're not a good fit for man pages.)

What are we doing here then? What exactly goes into the docs vs just on the website?

zuiderkwast:
> > Even if we my want such guides on the website in the future, I doubt they belong in the doc repo. (They're not a good fit for man pages.)
> 
> What are we doing here then? What exactly goes into the docs vs just on the website?

The border between docs and other content is a bit fuzzy.

Anyway, these guides seemed to be just copy-paste the client libs' README + boilerplate. It's better to link to the clients' README files directly to avoid the maintenance of this IMHO.

We can add always add back content later, but for Valkey clients in that case.

madolson:
OK, that's reasonable. I would like some more in depth tutorials, but they need not be on the website.

PR #22, opened 2024-04-09
Remove version details and history from doc files

Shivshankar-Reddy:
Removed the version details and history block from doc files.

PR #85, opened 2024-05-04
Add information about server domain object for LUA

madolson:
Documents the behavior for the new server top level domain object and the new Lua fields. Also updates all of the examples to use the server object. Also includes some cleanup to references to "Valkey serialization protocol" which doesn't exist. 

There will be some minimal conflicts with https://github.com/valkey-io/valkey-doc/pull/22, as it attempts to remove the versioning references which this file also does for the files that were touched.


topics/lua-api.md
```diff
@@ -502,9 +506,8 @@ You can use the following flags and instruct the server to treat the scripts' ex
 
 Please refer to [Function Flags](functions-intro.md#function-flags) and [Eval Flags](eval-intro.md#eval-flags) for a detailed example.
 
-### <a name="redis.redis_version"></a> `redis.REDIS_VERSION`
+### <a name="server.server_version"></a> `server.SERVER_VERISON`
 
-* Since version: 7.0.0
```
zuiderkwast, line 507: I think the version information is important. I disagree with the PR deleting all this.

Users may write Lua scripts that are need to run on a range of versions of Valkey and Redis OSS, so depending on which versions they need to support, they need to know what they can use. Let's say they're writing some open source lib or tool or so.

For things that aren't yet released, it's even more important, but this one is new in Valkey 7.2.5, right?

Somewhere in the docs, we can write that versions < 7.2.5 always refer to Redis OSS and >= 7.2.5 refer to Valkey versions.

topics/lua-api.md
```diff
@@ -515,9 +518,20 @@ The reply's format is `MM.mm.PP`, where:
 * **mm:** is the minor version.
 * **PP:** is the patch level.
 
-### <a name="redis.redis_version_num"></a> `redis.REDIS_VERSION_NUM`
+### <a name="server.redis_version"></a> `server.REDIS_VERSION`
+
+* Available in scripts: yes
```
zuiderkwast, line 523: If called as `redis.REDIS_VERSION`, it came in 7.0.0.

```suggestion
### <a name="server.redis_version"></a> `server.REDIS_VERSION`

* Since version: 7.0.0
* Available in scripts: yes
```

topics/lua-api.md
```diff
@@ -617,7 +643,7 @@ As you can see, the float value of _3.333_ gets converted to an integer _3_, the
 
 ### RESP3 to Lua type conversion
 
-[RESP3](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md) is a newer version of the [Valkey Serialization Protocol](protocol.md).
+[RESP3](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md) is a newer version of the [Redis Serialization Protocol](protocol.md).
```
zuiderkwast, line 646: RESP3 is documented in protocol.md nowdays.

topics/lua-api.md
```diff
@@ -79,11 +79,15 @@ The only libraries that Valkey ships with and that you can use in scripts are li
...
 
-The _redis_ singleton is an object instance that's accessible from all scripts.
+The _server_ singleton is an object instance that's accessible from all scripts.
 It provides the API to interact with Valkey from scripts.
-Its description follows [below](#redis_object).
+Its description follows [below](#server_object).
+
+**Note:**
+For compatibility with Redis, Valkey also exposes a _redis_ top level object, that exposes the exact same set of APIs as the _server_ object.
+Valkey does not intend to drop compatibility for this _redis_ API, but it is recommended to use the _server_ object for newly developed scripts.
```
zuiderkwast, line 90: If we keep the "Since" information, we can explain here that the `server` object exists since 7.2.5. Earlier versions (Redis OSS) only support the `redis` object.

zuiderkwast (commented):
Good. I just think we should keep version information. Let's discuss that.

topics/lua-api.md
madolson: I'm not sure how strongly I feel about that. There are some very old versions of Redis that have weird behavior, are we going to document all of that behavior? Is there a single issue where we are discussing this? Should we add it to the contributor meeting we have on Monday?

> Users may write Lua scripts that are need to run on a range of versions of Valkey and Redis OSS, so depending on which versions they need to support, they need to know what they can use. Let's say they're writing some open source lib or tool or so.

I don't think it's our job to detail all of the historical behavior changes in Redis. I am documenting all of the "breaking" changes, and intend to write a single page including everything that could break if you are moving from Redis to Valkey.

zuiderkwast: Old versions, please delete. The only ones (IMO) that may be of interest are the still supported versions, i.e. 6.2, 7.0 and 7.2.

It's been discussed in random PRs. There's no single place AFAIK.

> I don't think it's our job to detail all of the historical behavior changes in Redis.

That's one way to look at it.

Here's another: If you consider Valkey the continuation of Redis OSS that just changed its name, then it does make sense to keep historical info. We're still the same project, so version history is as useless or useful as it would be without a name change.

> intend to write a single page including everything that could break if you are moving from Redis to Valkey.

Good, but I think that's release notes material. That's what people should review when updating. If people don't read the release notes, then maybe they're not accessible enough? We should link to them on the download page next to each version.

madolson: > Old versions, please delete. The only ones (IMO) that may be of interest are the still supported versions, i.e. 6.2, 7.0 and 7.2.

Ok. I'd commit to keeping information about versions 7-6.2 for historical purposes. We can document we drop references from versions of Redis before that point.

> Good, but I think that's release notes material. That's what people should review when updating. If people don't read the release notes, then maybe they're not accessible enough? We should link to them on the download page next to each version.

I was going to do it as a one-off to build trust for folks to migrate. I agree generally that is for release notes, but it will be confusing initially as folks move from Redis -> valkey.

topics/lua-api.md
```diff
@@ -528,6 +542,18 @@ The reply is a hexadecimal value structured as `0x00MMmmPP`, where:
 * **mm:** is the minor version.
 * **PP:** is the patch level.
 
+### <a name="server.redis_version_num"></a> `server.REDIS_VERSION_NUM`
+
+* Available in scripts: yes
```
zuiderkwast, line 547:
```suggestion
* Since version: 7.0.0 (if accessed as `redis.REDIS_VERSION_NUM`)
* Available in scripts: yes
```

topics/lua-api.md
```diff
@@ -539,7 +565,7 @@ that reply is automatically converted to Valkey' protocol.
 Put differently; there's a one-to-one mapping between Valkey' replies and Lua's data types and a one-to-one mapping between Lua's data types and the [Valkey Protocol](protocol.md) data types.
 The underlying design is such that if a Valkey type is converted into a Lua type and converted back into a Valkey type, the result is the same as the initial value.
 
-Type conversion from Valkey protocol replies (i.e., the replies from `server.call()` and `server.pcall()`) to Lua data types depends on the Valkey Serialization Protocol version used by the script.
+Type conversion from Valkey protocol replies (i.e., the replies from `server.call()` and `server.pcall()`) to Lua data types depends on the Redis Serialization Protocol version used by the script.
```
zuiderkwast, line 568:
```suggestion
Type conversion from Valkey replies (i.e. the replies from `server.call()` and `server.pcall()`) to Lua data types depends on the RESP protocol version used by the script.
```

zuiderkwast (approved):

PR #89, opened 2024-05-07
Fix return message of HSET command in HSCAN NOVALUES.

CharlesChen888:
Should return `(integer) 2`, rather than `OK`.


zuiderkwast (approved):
Thanks!

PR #94, opened 2024-05-23
Update release guidance for Valkey

madolson:
A fairly extensive rewrite of the Valkey release and versioning documentation from our docs. Generally made two significant changes.

1. Made much clearer guidance around semantic versioning and how we follow it.
2. Made more explicit the support model for how long we maintain a release.

@valkey-io/core-team and @valkey-io/contributors, PTAL.


topics/releases.md
```diff
@@ -1,144 +1,98 @@
...
-A new major version is planned for release once a year.
+The Valkey community strives to release a stable major version once a year.
+Stable minor versions are created as needed in between major releases, and we aim to release at least one minor version a year.
 
-Generally, every major release is followed by a minor version after six months.
+### Release candidate
 
-Patches are released as needed to fix high-urgency issues, or once a stable
-version accumulates enough fixes to justify it.
+New minor and major versions of Valkey begin by branching off the `unstable` branch as an initial release candidate branch, which take the form *major.minor.patch-R#*.
```
enjoy-binbin, line 62:
```suggestion
New minor and major versions of Valkey begin by branching off the `unstable` branch as an initial release candidate branch, which take the form *major.minor.patch-RC#*.
```

enjoy-binbin (approved):

topics/releases.md
```diff
@@ -1,144 +1,98 @@
 ---
-title: "Valkey release cycle"
-linkTitle: "Release cycle"
+title: "Valkey releases and versioning"
+linkTitle: "Valkey releases"
 weight: 4
-description: How are new versions of Valkey released?
+description: How new versions of Valkey released and supported
```
zuiderkwast, line 5:
```suggestion
description: How new versions of Valkey are released and supported
```

topics/releases.md
```diff
@@ -1,144 +1,98 @@
...
-backward compatibility.
+Valkey stable releases will follow generally `major.minor.patch` [semantic versioning schema](https://semver.org/).
+We follow semantic versioning to provide explicit guarantees regarding backward compatibility.
 
-### Patch-Level versions
+### Patch versions
 
-Patches primarily consist of bug fixes and very rarely introduce any
-compatibility issues.
+PATCH versions are released with backwards compatible bug fixes and should not introduce new features.
```
zuiderkwast, line 22: Uppercase as emphasis? It looks like a command. :)

```suggestion
*Patch* versions are released with backwards compatible bug fixes and should not introduce new features.
```

topics/releases.md
```diff
@@ -1,144 +1,98 @@
...
 
-Upgrading from a previous patch-level version is almost always safe and
-seamless.
+Upgrading from a previous patch version should be safe and seamless.
+It should be safe to run a Valkey cluster with servers running on different patch versions.
```
zuiderkwast, line 25: Speaking of cluster, we can also mention replication from/to?

topics/releases.md
```diff
@@ -1,144 +1,98 @@
...
+MINOR version are released with new functionality that is added in a backward compatible manner.
+Examples of new functionality include new commands, info fields, or configuration parameters.
+
+Upgrading from a previous minor version should be safe, and will not introduce incompatibilities between servers in the cluster.
 
-Upgrading between minor versions does not introduce any application-level
-compatibility issues.
+**NOTE:** Minor releases may include new commands and data types that can introduce incompatibility between servers in the cluster, but users need to opt-in to these features to cause this type of incompatibility.
+For this reason, it is not recommended to run a Valkey cluster with servers running on different minor versions.
+Users should also avoid new features until all servers in the cluster have been upgrades.
```
zuiderkwast, line 38:
```suggestion
Users should also avoid new features until all servers in the cluster have been upgraded.
```

topics/releases.md
```diff
@@ -1,144 +1,98 @@
...
 ### Major versions
 
-Major versions introduce new capabilities and significant changes.
+MAJOR versions are released with significant functionality that may break backwards compatibility or alter key performance characteristics.
+Examples of significant functionality include altering the behavior of an existing command, removing previously deprecated commands, changing the default value of configs, and significant refactoring for performance improvements.
 
-Ideally, these don't introduce application-level compatibility issues.
+Upgrading from a previous major version is intended to be safe, but should be approached with caution. 
+You should carefully read the release notes before performing a major version upgrade.
+Major version upgrades do not guarantee backwards compatibility, which means you should always upgrade replicas before upgrading primaries in order to ensure data consistency.
```
zuiderkwast, line 50: How to do rolling upgrade in a cluster or sentinel is described somewhere in the docs, right? Can we link to it?

topics/releases.md
zuiderkwast: The release candidate *branch* is just *major.minor*.

The release candidate *tags* are with lowercase "rc" IIRC. I don't know if it's clear (in all parts of the world) that '#' is a place-holder for a digit. Can we use N or something instead, like "_X.Y.Z_-**rc**_N_", or by example *8.0.0-rc1*, *8.0.0-rc2*, etc.

The patch number is always 0 in release canditates?

topics/releases.md
```diff
@@ -1,144 +1,98 @@
...
 The latest stable release is always fully supported and maintained.
 
-Two additional versions receive maintenance only, meaning that only fixes for
-critical bugs and major security issues are committed and released as patches:
+The Valkey community will provide maintanence support, providing patch releases for bug fixes and all security fixes, for 3 years from when a version was first released.
+
+The Valkey community will also provide extended security security support for the latest minor version of each major version for 5 years from when a version was first released.
+The Valkey community will only backport security issues we believe to be possible to exploit, which will be up to the discretion of the TSC.
+
+For contacting the core team on sensitive matters and security issues, please see [SECURITY.md](https://github.com/valkey-io/valkey/blob/unstable/SECURITY.md).
```
zuiderkwast, line 83: Here, "TCS" and "core team" are used as synonyms, but I don't know if all readers know that.

topics/releases.md
```diff
@@ -1,144 +1,98 @@
...
+| -- | -- | -- | -- |
+| 7.2 | 2024-04-16 | 2027-04-16 | 2029-04-16 |
 
-Once version 3.0.0 replaces 2.2 as the latest stable, versions 2.0 and 2.2 are
-maintained, whereas version 1.x reaches its end of life.
+## Unstable tree
 
-This process repeats with version 3.2.0, after which only versions 2.2 and 3.0
-are maintained.
+The unstable development tree of Valkey is located in the `unstable` branch in the [Valkey GitHub repository](https://github.com/valkey-io/valkey).
```
zuiderkwast, line 93:
```suggestion
The development tree of Valkey is located in the `unstable` branch in the [Valkey GitHub repository](https://github.com/valkey-io/valkey).
```

zuiderkwast (commented):
Very good.

Shall we continue using only even numbers for minor versions? If yes, shall we mention this anywhere?

topics/releases.md
```diff
@@ -1,144 +1,98 @@
...
-
-### Stable tree
-
-Once development has ended and the frequency of critical bug reports for the
-release candidate wanes, it is ready for the final release. At this point, the
-release is marked as stable and is released with "0" as its patch-level
-version.
+All Valkey releases are published in the [Valkey GitHub repository](https://github.com/valkey-io/valkey/releases).
 
 ## Versioning
```
madolson, line 15: We should be more clear about the backwards compatibility about:
1. RDB Versioning
2. API versioning
3. Replication protocol
4. Cluster protocol.

madolson: Modules?

topics/releases.md
madolson: I couldn't find an example of it. It's on the list of stuff we should document better. https://github.com/valkey-io/valkey-doc/issues/115

topics/releases.md
madolson: I didn't follow this comment.

topics/releases.md
```diff
@@ -1,144 +1,109 @@
...
-the release's purpose is collecting the public's feedback.
+When discussing compatibility, we refer to the following API contracts:
 
-Subsequent release candidates are released every three weeks or so, primarily
-for fixing bugs. These may also add new features and introduce changes, but at
-a decreasing rate and decreasing potential risk towards the final release
-candidate.
+1. Valkey commands including their inputs, outputs, and defined behavior
+2. The functions and APIs that can be executed from a Lua script
+3. The RDB version
```
PingXie, line 24: AOF is the other persistence mechanism we support. It has its own on-disk format so it would be worth mentioning here too.

topics/releases.md
PingXie: I think @zuiderkwast is saying we support replicating from the lower version of Valkey/Redis to a higher version but no guarantee if you replicate from a higher version of Valkey to a lower version (because RDB version might get bumped - separate discussion).

topics/releases.md
```diff
@@ -1,144 +1,109 @@
...
+### Minor versions
 
-Patches primarily consist of bug fixes and very rarely introduce any
-compatibility issues.
+*Minor* versions are released with new functionality that is added in a backward compatible manner.
+Examples of new functionality include new commands, info fields, or configuration parameters.
 
-Upgrading from a previous patch-level version is almost always safe and
-seamless.
+Upgrading from a previous minor version should be safe, and will not introduce incompatibilities between servers in the cluster.
```
PingXie, line 43:
```suggestion
Upgrading from a previous minor version should be safe, and will not introduce incompatibilities between servers in the cluster if default server configurations are not changed.
```

topics/releases.md
```diff
@@ -1,144 +1,109 @@
...
 
-New features and configuration directives may be added, or default values
-changed, as long as these don’t carry significant impacts or introduce
-operations-related issues.
+**NOTE:** Minor releases may include new commands and data types that can introduce incompatibility between servers in the cluster, but users need to opt-in to these features to cause this type of incompatibility.
+For this reason, it is not recommended to run a Valkey cluster with servers running on different minor versions.
+Users should avoid new features until all servers in the cluster have been upgraded.
 
-### Minor versions
+Commands may also be deprecated in minor versions.
```
PingXie, line 49:
```suggestion
Commands may also be marked as `deprecated` in minor versions.
```

topics/releases.md
```diff
@@ -1,144 +1,109 @@
...
+If a command is deprecated, a replacement command or an alternative to using the command will be defined in the same minor version.
```
PingXie, line 50:
```suggestion
Deprecated commands are not removed, but a replacement command or an alternative to using the command will be defined in the same minor version.
```

topics/releases.md
```diff
@@ -1,144 +1,109 @@
...
-compatibility issues.
+*Major* versions are released with significant functionality that may break backwards compatibility or alter key performance characteristics.
+Examples of significant functionality includes altering the behavior of an existing command, removing previously deprecated commands, changing the default value of configs, and significant refactoring for performance improvements.
 
-Minor releases may include new commands and data types that introduce
-operations-related incompatibilities, including changes in data persistence
-format and replication protocol.
+Upgrading from a previous major version is intended to be safe, but should be approached with caution. 
+You should carefully read the release notes before performing a major version upgrade.
+Major version upgrades do not guarantee backwards compatibility, which means you should always upgrade replicas before upgrading primaries in order to ensure data consistency.
```
PingXie, line 59: >Major version upgrades do not guarantee backwards compatibility, which means you should always upgrade replicas before upgrading primaries in order to ensure data consistency.

If we don't guarantee "backwards compatibility", how do we guarantee users can always replicate a older version of Valkey to the latest major version? I wonder if we should take the similar approach like in the minor version upgrade case where potentially incompatible changes are gated behind a server config?

topics/releases.md
```diff
@@ -1,144 +1,109 @@
...
+Stable minor versions are created as needed in between major releases, and we aim to release at least one minor version a year.
 
-## Release schedule
+### Release candidate
 
-A new major version is planned for release once a year.
+New minor and major versions of Valkey begin by branching off the `unstable` branch as an initial release candidate branch with a name that takes the form of `*major.minor*`, example `7.2`.
+The first release candidate, or rc1, is released once it can be used for development purposes and for testing the new version.
+Release candidate versions will start with a patch version of "0" and will take the form `*major.minor.patch-rcN*`, example `7.2.0-rc1` followed by `7.2.0-rc2`.
+At this stage, most of the new features and changes in the new version are ready for review, and the version is released for the purpose of collecting the public's feedback.
```
PingXie, line 74:
```suggestion
At this stage, most of the new features and changes in the new version are ready for review, and the version is released for the purpose of collecting public feedback.
```

topics/releases.md
```diff
@@ -1,144 +1,109 @@
...
 
-As a rule, older versions are not supported as we try very hard to make the
-API mostly backward compatible.
+The latest stable release is always fully supported and maintained.
 
-Upgrading to newer versions is the recommended approach and is usually trivial.
+The Valkey community will provide maintanence support, providing patch releases for bug fixes and all security fixes, for 3 years from when a version was first released.
 
-The latest stable release is always fully supported and maintained.
+The Valkey community will also provide extended security security support for the latest minor version of each major version for 5 years from when a version was first released.
```
PingXie, line 90:
```suggestion
The Valkey community will also provide extended security support for the latest minor version of each major version for 5 years from when a version was first released.
```

topics/releases.md
```diff
@@ -1,144 +1,109 @@
...
+Patches are released as needed to fix high-urgency issues, or once a stable version accumulates enough fixes to justify it.
 
 ## Support
 
-As a rule, older versions are not supported as we try very hard to make the
-API mostly backward compatible.
+The latest stable release is always fully supported and maintained.
 
-Upgrading to newer versions is the recommended approach and is usually trivial.
+The Valkey community will provide maintanence support, providing patch releases for bug fixes and all security fixes, for 3 years from when a version was first released.
```
PingXie, line 88: I think we should also clarify for which minor version we will provide "maintenance support" because below we have been very explicit that only the last minor version of each major version will receive security support.

IMO, not providing security support for other minor versions in a major version makes the "maintenance support" for non-last minor versions a moot point because users need both. So I think either we should provide both support for every minor version or align maintenance support with security support so only the last minor version of each major version is considered (my preference)

```suggestion
The Valkey community will provide maintenance support, providing patch releases for bug fixes and all security fixes, for 3 years from when a version was first released.
```

topics/releases.md
madolson: Yeah, good point

topics/releases.md
madolson: Ok, let me try to add a sentence to address this.

topics/releases.md
madolson: Assuming you want this to be emphasized, I'm going to build it insteadl.

topics/releases.md
madolson: > IMO, not providing security support for other minor versions in a major version makes the "maintenance support" for non-last minor versions a moot point because users need both.

This is definitely not true. If there is a bug in a feature you don't use, you won't see any reason to upgrade. It'll also be easier if we say we aren't going to fix it.  Many organizations are required to follow procedures for patching security issues. We have historically had about 2 security CVE fixes per year in Redis since we have so few dependencies. 

> So I think either we should provide both support for every minor version or align maintenance support with security support so only the last minor version of each major version is considered (my preference)

Everyone I've pitched this proposal to, doesn't like it. Since it will push end users to wait until LTS versions are released and then they'll upgrade, at which point, I would prefer we just launch a single major version every year with features.

Second question, what would be the support of a minor version? I don't think it should just be until the next minor is released, that is **very** uncommon.

topics/releases.md
madolson: I needed to rework this section, since we have some odd guarantees here.

topics/releases.md
madolson: I interpreted your suggestion as if you are using default configuration, behavior won't change. The current wording you have meant to me that the behavior won't change if **we** don't change default configurations, which we don't do in a minor version.

wordlist
```diff
@@ -635,7 +635,7 @@ race:france
 race:italy
 race:usa
 radix
-rc
+rc[0-9]*
```
PingXie, line 638: should it be a `+`?
```suggestion
rc[0-9]+
```

topics/releases.md
PingXie: > If there is a bug in a feature you don't use, you won't see any reason to upgrade. 

Good point.

>If there is a security issue, you'll organization will probably make you apply it.

I read the latest revision as saying all versions (minor and major) are fully supported for up to 3 years and between year 4 and year 5, inclusive, only last minor version of each major version is supported for security. Is this understanding correct?

Here is an example, and assuming we tick-tock release every 6 months: 7.2, 8.0, 8.2, 9.0, 9.2, 10.0, 10.2, 11.0, 11.2, 12.0, and 12.2.

By the time we release 12.2, 7.2 turns 5 and 9.2 turns 3. We then have a CVE fix and a critical bug fix applicable to all these versions. With this release policy, are we saying, versions between 9.2 and 12.2, inclusive, will receive both fixes, which is 7 versions in total but 7.2 and 8.2 will receive the CVE fix only?

PingXie (commented):
LGTM. I have just one last clarification question to validate my understanding.

topics/releases.md
zuiderkwast: I meant just primary-replica of different patch versions, not necessarily using cluster.

```suggestion
It should be safe to run a primary-replica pair or a Valkey cluster with servers running on different patch versions.
```

topics/releases.md
```diff
@@ -1,144 +1,111 @@
...
+The Valkey community strives to release a stable major version once a year.
+Stable minor versions are created as needed in between major releases, and we aim to release at least one minor version a year.
 
-## Release schedule
+### Release candidate
 
-A new major version is planned for release once a year.
+New minor and major versions of Valkey begin by branching off the `unstable` branch as an initial release candidate branch with a name that takes the form of `*major.minor*`, example `7.2`.
+The first release candidate, or rc1, is released once it can be used for development purposes and for testing the new version.
+Release candidate versions will start with a patch version of "0" and will take the form `*major.minor.patch-rcN*`, example `7.2.0-rc1` followed by `7.2.0-rc2`.
```
zuiderkwast, line 75: `` `*major.minor*` `` is rendered as `*major.minor*` with the stars visible. I think you may want to put the backticks inside the stars?

`` *`major.minor`* `` rendered as *`major.minor`*.

Another option is to just avoid backticks here. A version number isn't really code. We could combine bold and italics like

_major_**.**_minor_**.**_patch_**-rc**_N_ (where bold is fixed content and italics is variable content)

```suggestion
New minor and major versions of Valkey begin by branching off the `unstable` branch as an initial release candidate branch with a name that takes the form of *`major.minor`*, example `7.2`.
The first release candidate, or rc1, is released once it can be used for development purposes and for testing the new version.
Release candidate versions will start with a patch version of "0" and will take the form *`major.minor.patch-rcN`*, example `7.2.0-rc1` followed by `7.2.0-rc2`.
```

wordlist
zuiderkwast: "rc" without numbers was already in the wordlist, so it may be good to allow zero digits here to keep "rc" a valid word.

zuiderkwast (approved):
Good. (Just some nits.)

wordlist
madolson: Yeah, * was to cover both `rc` and `rc1`.

topics/releases.md
madolson: I understand now! I can add that.

PingXie (approved):
LGTM.

topics/releases.md
```diff
@@ -1,141 +1,108 @@
...
+Examples of new functionality include new commands, info fields, or configuration parameters.
 
-Upgrading from a previous patch-level version is almost always safe and
-seamless.
+Upgrading from a previous minor version should be safe, and will not introduce incompatibilities between servers in the cluster when default server configurations are used.
 
-New features and configuration directives may be added, or default values
-changed, as long as these don’t carry significant impacts or introduce
-operations-related issues.
+**NOTE:** Minor releases may include new commands and data types that can introduce incompatibility between servers in the cluster, but users need to opt-in to these features to cause this type of incompatibility.
```
soloestoy, line 43: Here's an issue. Generally, when we introduce a new data type, the `rdb_version` is also updated accordingly. So, even if users haven't used the new data type, the newer version nodes in the cluster would already be utilizing the updated `rdb_version`. In this scenario, if data is migrated from a newer version to an older version, errors would occur.

zuiderkwast: Yes. If we want to be forward-compatible, which I think is a good idea, we need to support writing RDB in an old version and a new version. We only need to be forward compatible one version, I think.

HLD: The first time a new datatype is used, we set a global flag. When we write the RDB, we check the flag. If the flag is set, we use the new RDB version. If the flag is not set, we use the older RDB version.

zuiderkwast: @soloestoy the long term solution is AOF without RDB? valkey-io/valkey#59

PR #95, opened 2024-05-24
Fix spellcheck

### LINKED ISSUE #93: Fix spellchecker
The spellchecker works as described in the README.md, but there are some problems:

* The spellchecker ignores text within backticks, but many command arguments and similar are not written in backticks, so they are reported as spelling errors. Solution: Add backticks for these.
* Our config file is `.spellcheck.yml` but according to https://www.npmjs.com/package/spellchecker-cli it needs to be called `.spellcheckerrc.yml`.
* Some words need to be added to the wordlist. Regexes can be used. For example `ValkeyModule.*`.
* Probably there are some actual spelling errors that need to be corrected.

zuiderkwast:
The wordlist is modified to accept all words in UPPERCASE, since those are the ones causing the most trouble and are often command arguments or acronyms not known by the spellchecker. Some more regexes are added to the wordlist.

The CI job is modified to use another action.

A bunch of actual spelling errors are fixed.

Fixes #93


madolson (approved):

zuiderkwast:
I'll update README too. With a correctly named config file, `spellcheck` without arguments in the way to run it locally.

PR #144, opened 2024-06-17
Mention that eviction policies use approximated randomized algorithms

enjoy-binbin:
We clearly indicate in the documentation that LRU / LFU are using
approximated randomized algorithms. Some people think volatile-ttl
is accurate, i suppose adding this sentence will be helpful, the text
is from valkey.conf

### DIFF
--- a/topics/lru-cache.md
+++ b/topics/lru-cache.md
@@ -57,7 +57,7 @@ The following policies are available:
 
 The policies **volatile-lru**, **volatile-lfu**, **volatile-random**, and **volatile-ttl** behave like **noeviction** if there are no keys to evict matching the prerequisites.
 
-Both **LRU**, **LFU** and **volatile-ttl** are implemented using approximated randomized algorithms.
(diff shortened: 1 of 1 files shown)


topics/lru-cache.md
```diff
@@ -57,6 +57,8 @@ The following policies are available:
 
 The policies **volatile-lru**, **volatile-lfu**, **volatile-random**, and **volatile-ttl** behave like **noeviction** if there are no keys to evict matching the prerequisites.
 
+Both **LRU**, **LFU** and **volatile-ttl** are implemented using approximated randomized algorithms.
```
zuiderkwast, line 60:
```suggestion
**LRU**, **LFU** and **volatile-ttl** are implemented using approximated randomized algorithms.
```

zuiderkwast (approved):
Good.

Why is the title bad? Too long?

Shorter idea: "Mention that eviction policies use approximated randomized algorithms"

I think the word "both" means "each of the two", but here we are talking about three things.

Issue #140, opened 2024-05-31
`/topics/cluster-spec.md` has outdated info


stockholmux:
In pre-publishing review (#91), I found the following issues:

> What is described in this document is implemented in Redis OSS 3.0 or greater.

Drop this reference to an ancient version.

> CLUSTER MEET ip port

Use of code backticks for inline code.

> Copyright 2010 Salvatore Sanfilippo (adapted to Valkey coding style)

This should probably revert back to 'redis' since Salvatore didn't adapt it to Valkey.

All code blocks should move to backticks/syntax highlighter for readability.

Finally, looks like there are 196 references that need to be updated to 'primary'.

zuiderkwast:
I'm removing the reference to Redis OSS 3.

> All code blocks should move to backticks/syntax highlighter for readability.

There's no usable language to use for syntax highlighting of Valkey commands like `CLUSTER MEET ip port`. If there would be a language we could mark it as, I'd buy the change to backticks, but without it, I'm going to ignore this comment. We're not changing just for the sake of changing.

stockholmux:
It's not a change for the sake of changing.

`CLUSTER MEET ip port` doesn't render properly as a code block in either GitHub nor on the website.

<img width="797" alt="Screenshot 2024-06-20 at 1 47 18 PM" src="https://github.com/valkey-io/valkey-doc/assets/1152927/228fdf23-de11-4dff-b14e-be9aacf99af6">
<img width="969" alt="Screenshot 2024-06-20 at 1 47 12 PM" src="https://github.com/valkey-io/valkey-doc/assets/1152927/7c224f9b-c759-44cd-a231-ec6e732f33fd">

zuiderkwast:
Do you mean it's indented less than 4 spaces? Because there's no difference between


```
CLUSTER MEET ip port
```

and

    CLUSTER MEET ip port

zuiderkwast:
It's under a bullet list item so it'd need to be indented more. Or switch to backticks to avoid the problem of nested indentation.

The crap about syntax highlighing is what made me disregard the entire comment as trolling.

stockholmux:
I am (generally) not trolling. I actually think it should be fixed as in #149.

The use of the code block here is super inconsistent with the rest of the document.

zuiderkwast:
Hehe, I know! 😄 But syntax highlighting for valkey commands and other pseudo code is still hard to take serious, unless you have a specific language it can be tagged as. 😉

[...]

PR #149, opened 2024-06-20
Resolves ambiguous markdown in `cluster-spec.md`

stockholmux:
Fixes issue (#140) with `CLUSTER MEET ip port` rendering improperly.


zuiderkwast (approved):
Inline code, yes, it's better.

PR #155, opened 2024-07-05
Doc review fixes part 2

### LINKED ISSUE #113: `/topics/latency.md` is out dated information an references non existent things
In pre-publishing review (#91), the following issues were discovered

> This is an example run under Linux 3.11.0 running on an entry level server.

Linux 3.11.0 is from 2013-ish and is long out of support. It would be good to modernize this section with a more common kernel in 2024.

> new types of EC2 HVM based instances are much better with fork times

I don't think these would be considered new. Additionally there is an entire list of machines and their performance; I think most of these machines are from about a decade ago and their performance provides little value in 2024.

> Valkey Google Group

(two mentions). No such google group exists.

### LINKED ISSUE #114: `/topics/lru-cache.md` has old references
In pre-publishing review (#91), I discovered the following issue with `/topics/lru-cache.md`:

> However, since Redis OSS 3.0 the algorithm was improved to also take a pool of good candidates for eviction. 

And then proceeds to illustrate how Redis 2.8 and 3.0 compare. These are very old, irrelevant versions.

> The new LFU mode... Starting with Redis OSS 4.0....

These sections should be refactored to talk about the current state and now provide version comparisons for years old versions of another project.

### LINKED ISSUE #115: Create a detailed doc reference for doing rolling upgrades
For standalone, sentinel, and cluster mode, we should have crisp documentation explaining how to do a rolling upgrade in various configurations.

### LINKED ISSUE #117: `/topics/mass-insertion.md` has protocol naming and code block issue(s)
In pre-publishing review (#91), I found the following problems with `mass-insertion.md`

> Generating Valkey Protocol

and

> The Valkey protocol

Should probably be 'RESP'

> The following Ruby function generates valid protocol:

The subsequent code block should be marked as Ruby.

### LINKED ISSUE #119: `/topics/memory-optimization.md` has out dated information and confusing headings
In pre-publishing review (#91), I found the following issues with `/topics/memory-optimization.md`

The names in the headings for `Valkey <= 6.2` and `Redis OSS >= 7.0` are swapped (and slightly confusing)

> This is the result against a 64 bit instance of Redis OSS 2.2:

Redis OSS 2.2 is not relevant in 2024

> WARNING: for this to work, make sure that in your valkey.conf you have something like this:
hash-max-zipmap-entries 256

I don't think this will work in Valkey since 7.0 switched to listpack

zuiderkwast:
Fixes #113, fixes #114, fixes #115, fixes #117, fixes #119.

memory-optimization.md, mass-insertion.md, lua-api.md, lfu-cache.md, latency.md

### DIFF
--- a/topics/latency.md
+++ b/topics/latency.md
@@ -75,7 +75,8 @@ or setup.
 
 We call this kind of latency **intrinsic latency**, and `valkey-cli`
 is able to measure it. This is an example run
-under Linux 3.11.0 running on an entry level server.
+under Linux 3.11.0 running on an entry level server around 2014.
+(This is old, but it illustrates how to measure this.)
 
 Note: the argument `100` is the number of seconds the test will be executed.
 The more time we run the test, the more likely we'll be able to spot
@@ -233,24 +234,13 @@ of a large memory chunk can be expensive.
 Fork time in different systems
 ------------------------------
 
-Modern hardware is pretty fast at copying the page table, but Xen is not.
-The problem with Xen is not virtualization-specific, but Xen-specific. For instance using VMware or Virtual Box does not result into slow fork time.
-The following is a table that compares fork time for different Valkey instance
-size. Data is obtained performing a BGSAVE and looking at the `latest_fork_usec` filed in the `INFO` command output.
+Modern hardware is pretty fast at copying the page table.
+So are modern hardware-assisted virtualized environments,
+but fork can be really slow in older virtualized environmants without hardware support.
+As of 2024, this is hardly a problem.
 
-However the good news is that **new types of EC2 HVM based instances are much
-better with fork times**, almost on par with physical servers, so for example
-using m3.medium (or better) instances will provide good results.
-
-* **Linux beefy VM on VMware** 6.0GB RSS forked in 77 milliseconds (12.8 milliseconds per GB).
-* **Linux running on physical machine (Unknown HW)** 6.1GB RSS forked in 80 milliseconds (13.1 milliseconds per GB)
-* **Linux running on physical machine (Xeon @ 2.27Ghz)** 6.9GB RSS forked into 62 milliseconds (9 milliseconds per GB).
-* **Linux VM on 6sync (KVM)** 360 MB RSS forked in 8.2 milliseconds (23.3 milliseconds per GB).
-* **Linux VM on EC2, old instance types (Xen)** 6.1GB RSS forked in 1460 milliseconds (239.3 milliseconds per GB).
-* **Linux VM on EC2, new instance types (Xen)** 1GB RSS forked in 10 milliseconds (10 milliseconds per GB).
-* **Linux VM on Linode (Xen)** 0.9GBRSS forked into 382 milliseconds (424 milliseconds per GB).
-
-As you can see certain VMs running on Xen have a performance hit that is between one order to two orders of magnitude. For EC2 users the suggestion is simple: use modern HVM based instances.
+You can measure the fork time for a Valkey instance by
+performing a BGSAVE and looking at the `latest_fork_usec` filed in the `INFO` command output.
 
 Latency induced by transparent huge pages
 -----------------------------------------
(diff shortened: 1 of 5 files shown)


topics/latency.md
```diff
@@ -233,24 +234,13 @@ of a large memory chunk can be expensive.
...
-* **Linux running on physical machine (Unknown HW)** 6.1GB RSS forked in 80 milliseconds (13.1 milliseconds per GB)
-* **Linux running on physical machine (Xeon @ 2.27Ghz)** 6.9GB RSS forked into 62 milliseconds (9 milliseconds per GB).
-* **Linux VM on 6sync (KVM)** 360 MB RSS forked in 8.2 milliseconds (23.3 milliseconds per GB).
-* **Linux VM on EC2, old instance types (Xen)** 6.1GB RSS forked in 1460 milliseconds (239.3 milliseconds per GB).
-* **Linux VM on EC2, new instance types (Xen)** 1GB RSS forked in 10 milliseconds (10 milliseconds per GB).
-* **Linux VM on Linode (Xen)** 0.9GBRSS forked into 382 milliseconds (424 milliseconds per GB).
-
-As you can see certain VMs running on Xen have a performance hit that is between one order to two orders of magnitude. For EC2 users the suggestion is simple: use modern HVM based instances.
+You can measure the fork time for a Valkey instance by
+performing a BGSAVE and looking at the `latest_fork_usec` filed in the `INFO` command output.
```
zuiderkwast, line 243:
```suggestion
performing a BGSAVE and looking at the `latest_fork_usec` field in the `INFO` command output.
```

topics/lua-api.md
```diff
@@ -306,33 +296,13 @@ Please refer to the [Data type conversion](#data-type-conversion) for more infor
...
-
-Consider a script that intersects two sets and stores the result in a temporary key with `SUNIONSTORE`.
-It then picks five random elements (`SRANDMEMBER`) from the intersection and stores (`SADD`) them in another set.
-Finally, before returning, it deletes the temporary key that stores the intersection of the two source sets.
-
-In this case, only the new set with its five randomly-chosen elements needs to be replicated.
-Replicating the `SUNIONSTORE` command and the `DEL`ition of the temporary key is unnecessary and wasteful.
+Script replication is no longer supported.
+The only replication mode available is script effects replication.
+This feature is only available in Redis OSS versions prior to 7.0 which supported and used verbatim script replication by default.
```
stockholmux, line 305: This is a little confusing: I couldn't parse what 'this' in the 3rd sentence, especially with the present tense. 

Maybe it should be something like: "Prior to Redis OSS 7.0, the default was verbatim script replication. After Redis OSS 7.0 (and Valkey), script effects replication is only replication mode available."

stockholmux (changes requested):
One nit and one suggestion

topics/latency.md
```diff
@@ -75,7 +75,8 @@ or setup.
 
 We call this kind of latency **intrinsic latency**, and `valkey-cli`
 is able to measure it. This is an example run
-under Linux 3.11.0 running on an entry level server.
+under Linux 3.11.0 running on an entry level server around 2014.
+(This is old, but it illustrates how to measure this.)
```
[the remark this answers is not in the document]
zuiderkwast: Yeah I tried on the laptop but I got pretty bad numbers. It needs a real machine. I opened #156.

Another option is to just not mention which hardware and kernel it was running on. The numbers could still make sense today, depending on hardware, OS and virtualized environment. A machine with lots of other loads can very well have these numbers today. What it does is actually "it will just try to measure the largest time the kernel does not provide CPU time to run to the valkey-cli process itself" and its on the microseconds scale. (I'd expect the difference to depend more on the scheduler, whether it's a real-time kernel, than anything else.)

The comment for the second run further down "The following is a run on a Linode 4096
instance running Valkey and Apache" we can change to something more generic like "in a virtualized environment sharing hardware with other loads".

topics/lua-api.md
zuiderkwast: Yeah, I see now the 3rd sentence contradicts the other sentences. I'll change it.

stockholmux (approved):
lgtm

zuiderkwast:
Thanks for reviewing!

PR #158, opened 2024-07-16
Update rdb save releated fileds to distinguish diskless RDB

enjoy-binbin:
Currently, these fields do not distinguish between diskless,
which leads to some inconsistency in the fields.

### DIFF
--- a/commands/info.md
+++ b/commands/info.md
@@ -418,14 +418,12 @@ Here is the meaning of all fields in the **persistence** section:
 *   `current_fork_perc`: The percentage of progress of the current fork process. For AOF and RDB forks it is the percentage of `current_save_keys_processed` out of `current_save_keys_total`.
 *   `current_save_keys_processed`: Number of keys processed by the current save operation
 *   `current_save_keys_total`: Number of keys at the beginning of the current save operation
-*   `rdb_changes_since_last_save`: Number of changes since the last dump
-*   `rdb_bgsave_in_progress`: Flag indicating a RDB save is on-going
-*   `rdb_last_save_time`: Epoch-based timestamp of last successful RDB save
-*   `rdb_last_bgsave_status`: Status of the last RDB save operation
-*   `rdb_last_bgsave_time_sec`: Duration of the last RDB save operation in
-     seconds
-*   `rdb_current_bgsave_time_sec`: Duration of the on-going RDB save operation
-     if any
+*   `rdb_changes_since_last_save`: Number of changes since the last RDB file save
+*   `rdb_bgsave_in_progress`: Flag indicating a RDB save is on-going, including a diskless replication RDB save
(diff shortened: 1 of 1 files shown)


commands/info.md
```diff
@@ -28,231 +28,231 @@ When no parameter is provided, the `default` option is assumed.
 
 ```valkey-cli
 127.0.0.1:6379> INFO
-# Server
```
enjoy-binbin, line 31: i tried editing via Github Web and it is still produced these changes. This seems to be a different save format (maybe it is the newline).

enjoy-binbin:
Maybe we can update the code, but i am not sure if we should update it. so just update the doc first.

commands/info.md
zuiderkwast: GitHub is owned by Microsoft so maybe they want to force DOS/Windows newline to everyone? :dizzy_face:

commands/info.md
```diff
@@ -418,14 +418,12 @@ Here is the meaning of all fields in the **persistence** section:
...
-*   `rdb_last_bgsave_time_sec`: Duration of the last RDB save operation in
-     seconds
-*   `rdb_current_bgsave_time_sec`: Duration of the on-going RDB save operation
-     if any
+*   `rdb_changes_since_last_save`: Number of changes since the last RDB file save
+*   `rdb_bgsave_in_progress`: Flag indicating a RDB save is on-going, including a diskless replication RDB save
+*   `rdb_last_save_time`: Epoch-based timestamp of last successful RDB file save
+*   `rdb_last_bgsave_status`: Status of the last RDB file save operation
+*   `rdb_last_bgsave_time_sec`: Duration of the last RDB save operation in seconds, including a diskless replication RDB save
+*   `rdb_current_bgsave_time_sec`: Duration of the on-going RDB save operation if any, including a diskless replication RDB save
```
zuiderkwast, line 426: Do these fields include diskless repliaction now, or is that the change you want to make to the code?

Open a code PR to discuss, I guess.

enjoy-binbin: `rdb_last_save_time` and `rdb_last_bgsave_status` does not include diskless replication.
what i am updating now is the current behavior.

RDB file means a bgsave or a disk-based replication

zuiderkwast: OK, so we can merge this. Can you fix the whitespace changes?

(I think INFO actually uses `\r\n` in the output.)

enjoy-binbin: I tried it before, and it changed no matter what I did...

(ide / editor / github web editor), both contains the whitespace changes.

i can try again later.

zuiderkwast: Hehe, ok. So this file has `\r\n` inside the INFO example but `n` everywhere else. :) I guess we don't need to preserve these invisible characters in the docs.

zuiderkwast (approved):

PR #159, opened 2024-07-25
Update cluster-slots to explain the deterministic ordering

roshkhatri:
Update cluster-slots docs
ref: https://github.com/valkey-io/valkey/pull/265

### DIFF
--- a/commands/cluster-slots.md
+++ b/commands/cluster-slots.md
@@ -1,4 +1,4 @@
-`CLUSTER SLOTS` returns details about which cluster slots map to which Valkey instances.The command response is deterministic across all nodes in a cluster and is ordered lexicographically.
+`CLUSTER SLOTS` returns details about which cluster slots map to which Valkey instances. 
 The command is suitable to be used by Valkey Cluster client libraries implementations in order to retrieve (or update when a redirection is received) the map associating cluster *hash slots* with actual nodes network information, so that when a command is received, it can be sent to what is likely the right instance for the keys specified in the command. 
 
 The networking information for each node is an array containing the following elements:
@@ -36,6 +36,8 @@ Each nested result is:
 Each result includes all active replicas of the master instance
 for the listed slot range. Failed replicas are not returned.
 
+The command response is deterministic across all nodes in a cluster, which means that if two nodes return the same response they have the same view of the cluster.
+Primaries are ordered by the slots they serve and then replicas are ordered lexicographically by the node-id they were assigned by the cluster. 
 The third nested reply is guaranteed to be the networking information of the master instance for the slot range.
 All networking information after the third nested reply are replicas of the master.
 
--- a/commands/cluster-slots.md
+++ b/commands/cluster-slots.md
@@ -34,7 +34,8 @@ Each nested result is:
   - ...continues until all replicas for this master are returned.
 
 Each result includes all active replicas of the master instance
-for the listed slot range.  Failed replicas are not returned.
+for the listed slot range. These active replicas are ordered deterministically.  
+Failed replicas are not returned.
 
 The third nested reply is guaranteed to be the networking information of the master instance for the slot range.
 All networking information after the third nested reply are replicas of the master.
--- a/commands/cluster-slots.md
+++ b/commands/cluster-slots.md
@@ -1,4 +1,4 @@
-`CLUSTER SLOTS` returns details about which cluster slots map to which Valkey instances. 
+`CLUSTER SLOTS` returns details about which cluster slots map to which Valkey instances.The command response is deterministic across all nodes in a cluster and is ordered lexicographically.
 The command is suitable to be used by Valkey Cluster client libraries implementations in order to retrieve (or update when a redirection is received) the map associating cluster *hash slots* with actual nodes network information, so that when a command is received, it can be sent to what is likely the right instance for the keys specified in the command. 
 
 The networking information for each node is an array containing the following elements:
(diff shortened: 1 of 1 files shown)


commands/cluster-slots.md
```diff
@@ -34,7 +34,8 @@ Each nested result is:
   - ...continues until all replicas for this master are returned.
 
 Each result includes all active replicas of the master instance
-for the listed slot range.  Failed replicas are not returned.
+for the listed slot range. These active replicas are ordered deterministically.  
+Failed replicas are not returned.
```
hpatro, line 38: I think we don't need a newline here.

commands/cluster-slots.md
```diff
@@ -34,7 +34,8 @@ Each nested result is:
   - ...continues until all replicas for this master are returned.
 
 Each result includes all active replicas of the master instance
-for the listed slot range.  Failed replicas are not returned.
+for the listed slot range. These active replicas are ordered deterministically.
```
hpatro, line 37: I think we could just mention that the command response is deterministic across all nodes in a cluster. WDYT ?

madolson: Also, should we clarify that they are ordered lexographically?

hwware: I think we should. Because  clients could check the result according to their ipaddress (becuase ipaddress as the first item in the result).

roshkhatri: >I think we could just mention that the command response is deterministic across all nodes in a cluster

Do you mean at the start of the doc? like this
```
`CLUSTER SLOTS` returns details about which cluster slots map to which Valkey instances. The command response is deterministic across all nodes in a cluster and is ordered lexicographically.
```

hpatro: I'm fine with the placement at the top.

commands/cluster-slots.md
```diff
@@ -34,7 +34,7 @@ Each nested result is:
   - ...continues until all replicas for this master are returned.
 
 Each result includes all active replicas of the master instance
-for the listed slot range.  Failed replicas are not returned.
+for the listed slot range. Failed replicas are not returned.
```
madolson, line 38:
```suggestion
for the listed slot range. Failed replicas are not returned.

The command response is deterministic across all nodes in a cluster, which means that if two nodes return the same response they have the same view of the cluster.
Instances are ordered by the slots they service and then replicas are ordered lexicographically by the node-id they were assigned by the cluster. 
```

commands/cluster-slots.md
```diff
@@ -1,4 +1,4 @@
-`CLUSTER SLOTS` returns details about which cluster slots map to which Valkey instances. 
+`CLUSTER SLOTS` returns details about which cluster slots map to which Valkey instances.The command response is deterministic across all nodes in a cluster and is ordered lexicographically.
```
madolson, line 1:
```suggestion
`CLUSTER SLOTS` returns details about which cluster slots map to which Valkey instances. 
```

madolson (commented):
I'm going to recommend we move it down below.

roshkhatri:
LGTM!

hpatro (approved):

commands/cluster-slots.md
```diff
@@ -34,8 +34,10 @@ Each nested result is:
   - ...continues until all replicas for this master are returned.
 
 Each result includes all active replicas of the master instance
-for the listed slot range.  Failed replicas are not returned.
+for the listed slot range. Failed replicas are not returned.
 
+The command response is deterministic across all nodes in a cluster, which means that if two nodes return the same response they have the same view of the cluster.
+Primaries are ordered by the slots they serve and then replicas are ordered lexicographically by the node-id they were assigned by the cluster. 
 The third nested reply is guaranteed to be the networking information of the master instance for the slot range.
 All networking information after the third nested reply are replicas of the master.
```
zuiderkwast, line 43: Here, we're jumping between describing the primary, the replicas and the slot ranges.

This section is about a *slot range*, which is each line in the nested reply. Looking at the whole text from "Each result includes all active replicas of the master instance for the listed slot range" until the example below, I would start with describing the ordering of the slot ranges (including the sentence that explains that if a shard has non-contiguous slots, then the networking information is repeated for each range) and then say what the primary for this slot range is and that the replicas for this slot range are ordered lexicographically by node-id.

Rough idea:

> Each result includes all active replicas of the primary instance for the listed slot range.
> Failing replicas are not returned.
>
> The slot ranges are returned in ascending order.
> If a cluster shard serves non-contiguous slots (e.g. 1-400,900,1800-6000) then primary and replica networking information results are duplicated for each slot range.
>
> The third nested reply is the networking information of the primary instance for the slot range.
> All networking information after the third nested reply are replicas of the primary.
> The replicas of each primary are ordered lexicographically by the node-id they were assigned by the cluster.

commands/cluster-slots.md
```diff
@@ -34,8 +34,10 @@ Each nested result is:
   - ...continues until all replicas for this master are returned.
 
 Each result includes all active replicas of the master instance
-for the listed slot range.  Failed replicas are not returned.
+for the listed slot range. Failed replicas are not returned.
 
+The command response is deterministic across all nodes in a cluster, which means that if two nodes return the same response they have the same view of the cluster.
```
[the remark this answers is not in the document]
zuiderkwast: I see that it was higher up earlier and was moved down by request by others. :) If it's not important enough to be at the top of the page, then maybe move it down further, to just before the example?

zuiderkwast (approved):
This is a small change that got stalled. Let's merge it. If we want to improve this document, we can do it later.

PR #167, opened 2024-08-26
Update cluster-spec with Valkey 8.0 cluster improvements

### LINKED ISSUE #76: [NEW] Document slot migration improvements
Many of these comments are perfect material for a doc PR.

`topics/cluster-tutorial.md` and maybe also `topics/cluster-spec.md`. If you start a doc PR, it can be useful also for us to read in parallel when reviewing this PR.

You could even use the good old style "Starting with Valkey 8.0, bla bla bla...".

File like https://github.com/valkey-io/valkey-doc/blob/main/topics/cluster-spec.md

_Originally posted by @zuiderkwast in https://github.com/valkey-io/valkey/pull/245#discussion_r1586608356_

PingXie:
Update cluster specification document to clarify shard migration process changes in Valkey 8.0 and include detailed descriptions of shard replication and failover mechanism improvements. 

Fix #76

### DIFF
--- a/commands/cluster-setslot.md
+++ b/commands/cluster-setslot.md
@@ -79,3 +79,12 @@ Notes:
   If the source node is informed before the destination node and the destination node crashes before it is set as new slot owner, the slot is left with no owner, even after a successful failover.
 * Step 6, sending `SETSLOT` to the nodes not involved in the resharding, is not technically necessary since the configuration will eventually propagate itself.
   However, it is a good idea to do so in order to stop nodes from pointing to the wrong node for the hash slot moved as soon as possible, resulting in less redirections to find the right node.
+* Starting from Valkey 8.0, `CLUSTER SETSLOT` is synchronously replicated to all healthy replicas
+  running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
+  If the replication fails, the primary does not execute the command, and the client receives a
+  `NOREPLICAS Not enough good replicas to write` error. Operators can retry the command or customize the
+  timeout using the `TIMEOUT` parameter to further increase the reliability of live reconfiguration:
+
+    CLUSTER SETSLOT slot [MIGRATING|IMPORTING|NODE] node-id [TIMEOUT timeout]
+
+  Here, `timeout` is measured in seconds, with 0 meaning to wait indefinitely.
--- a/commands/cluster-setslot.md
+++ b/commands/cluster-setslot.md
@@ -85,6 +85,8 @@ Notes:
   `NOREPLICAS Not enough good replicas to write` error. Operators can retry the command or customize the
   timeout using the `TIMEOUT` parameter to further increase the reliability of live reconfiguration:
 
-    CLUSTER SETSLOT slot [MIGRATING|IMPORTING|NODE] node-id [TIMEOUT timeout]
+  ```
+  CLUSTER SETSLOT slot [MIGRATING|IMPORTING|NODE] node-id [TIMEOUT timeout]
+  ```
 
   Here, `timeout` is measured in seconds, with 0 meaning to wait indefinitely.
--- a/topics/cluster-spec.md
+++ b/topics/cluster-spec.md
@@ -531,7 +531,7 @@ replicated to all replicas. This prevents loss of state if the primary fails aft
 
 Consider a scenario where the target primary node `B` is finalizing a slot migration.
 Before the `SETSLOT` command is replicated to its replica node `B’`, `B` might send a cluster `PONG`
-message to the source primary node `A`, promoting `A` to relinguish its ownership of the slot in question.
+message to the source primary node `A`, promoting `A` to relinquish its ownership of the slot in question.
 If `B` crashes right after this point, the replica node `B’`, which could be elected as the new primary,
 would not be aware of the slot ownership transfer without the synchronous replication of `SETSLOT`.
 This would leave the slot without an owner, leading to potential data loss and cluster topology inconsistency.
--- a/topics/cluster-spec.md
+++ b/topics/cluster-spec.md
@@ -403,7 +403,7 @@ without redirections, proxies or other single point of failure entities.
 A client **must be also able to handle -ASK redirections** that are described
 later in this document, otherwise it is not a complete Valkey Cluster client.
 
-### Live reconfiguration
+### Live resharding
 
 Valkey Cluster supports the ability to add and remove nodes while the cluster
 is running. Adding or removing a node is abstracted into the same
@@ -515,17 +515,17 @@ propagation of the new configuration across the cluster.
 
 #### Synchronous replication of `CLUSTER SETSLOT`
 
-Starting from Valkey 8.0, the `CLUSTER SETSLOT` command is synchronously replicated to all healthy replicas
-running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
+Starting from Valkey 8.0, the `CLUSTER SETSLOT` command is synchronously replicated if the replicas
+are running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
 If the replication fails, the primary does not execute the command, and the client receives a
 `NOREPLICAS Not enough good replicas to write` error. Operators can retry the command or customize the
-timeout using the `TIMEOUT` parameter to further increase the reliability of live reconfiguration:
+timeout using the `TIMEOUT` parameter to further increase the reliability of live resharding:
 
     CLUSTER SETSLOT slot [MIGRATING|IMPORTING|NODE] node-id [TIMEOUT timeout]
 
 Here, `timeout` is measured in seconds, with 0 meaning to wait indefinitely.
 
-Replicating the slot information prevents loss of state if the primary fails after executing the command.
+Synchronously replicating the slot information prevents loss of state if the primary fails after executing the command.
 Consider a scenario where the target primary node `B` is finalizing a slot migration.
 Before the `SETSLOT` command is replicated to its replica node `B’`, `B` might send a cluster `PONG`
 message to the source primary node `A`, promoting `A` to relinquish its ownership of the slot in question.
@@ -539,7 +539,7 @@ Starting from Valkey 8.0, Valkey clusters introduce the ability to elect a prima
 This behavior ensures that even when a shard is in the process of receiving its first slot,
 a primary can be elected. This prevents scenarios where there would be no primary available in the
 empty shard to handle redirected requests from the official slot owner,
-thereby maintaining availability during the live reconfiguration.
+thereby maintaining availability during the live resharding.
 
 ### ASK redirection
 
@@ -578,7 +578,7 @@ Slots migration is explained in similar terms but with different wording
 (for the sake of redundancy in the documentation) in the `CLUSTER SETSLOT`
 command documentation.
 
-Starting from Valkey 8.0, when the primary in either the source or target shard fails during live reconfiguration,
+Starting from Valkey 8.0, when the primary in either the source or target shard fails during live resharding,
 the primary in the other shard will automatically attempt to update its migrating/importing state to correctly pair
 with the newly elected primary. If this update is successful, the ASK redirection will continue functioning without
 requiring administrator intervention. In the event that slot migration fails, administrators can manually resume
--- a/topics/cluster-spec.md
+++ b/topics/cluster-spec.md
@@ -589,7 +589,7 @@ the interrupted slot migration by running the command `valkey-cli --cluster fix 
 
 Additionally, since Valkey 8.0, replicas are now able to return `ASK` redirects during slot migrations.
 This capability was previously unavailable, as replicas were not aware of ongoing slot migrations in earlier versions.
-It's worth noting that this change has already been documented in the READONLY command section.
+See the [READONLY](../commands/readonly.md) command.
 
 ### Client connections and redirection handling
 
--- a/topics/cluster-spec.md
+++ b/topics/cluster-spec.md
@@ -479,16 +479,6 @@ is that:
 * All queries about existing keys are processed by "A".
 * All queries about non-existing keys in A are processed by "B", because "A" will redirect clients to "B".
 
-Starting from Valkey 8.0, the `CLUSTER SETSLOT` command is synchronously replicated to all healthy replicas
-running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
-If the replication fails, the primary does not execute the command, and the client receives a
-`NOREPLICAS Not enough good replicas to write` error. Operators can retry the command or customize the
-timeout using the `TIMEOUT` parameter to further increase the reliability of live reconfiguration:
-
-    CLUSTER SETSLOT slot [MIGRATING|IMPORTING|NODE] node-id [TIMEOUT timeout]
-
-Here, `timeout` is measured in seconds, with 0 meaning to wait indefinitely.
-
 This way we no longer create new keys in "A".
 In the meantime, `valkey-cli` used during reshardings
 and Valkey Cluster configuration will migrate existing keys in
@@ -523,11 +513,37 @@ set the slots to their normal state again. The same command is usually
 sent to all other nodes to avoid waiting for the natural
 propagation of the new configuration across the cluster.
 
+#### Synchronous replication of `CLUSTER SETSLOT`
+
+Starting from Valkey 8.0, the `CLUSTER SETSLOT` command is synchronously replicated to all healthy replicas
+running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
+If the replication fails, the primary does not execute the command, and the client receives a
+`NOREPLICAS Not enough good replicas to write` error. Operators can retry the command or customize the
+timeout using the `TIMEOUT` parameter to further increase the reliability of live reconfiguration:
+
+    CLUSTER SETSLOT slot [MIGRATING|IMPORTING|NODE] node-id [TIMEOUT timeout]
+
+Here, `timeout` is measured in seconds, with 0 meaning to wait indefinitely.
+
+Synchronous replication is critical for maintaining cluster consistency during live reconfiguration.
+Before applying changes like slot ownership and migrating states to the primary, these must be fully
+replicated to all replicas. This prevents loss of state if the primary fails after executing the command.
+
+Consider a scenario where the target primary node `B` is finalizing a slot migration.
+Before the `SETSLOT` command is replicated to its replica node `B’`, `B` might send a cluster `PONG`
+message to the source primary node `A`, promoting `A` to relinguish its ownership of the slot in question.
+If `B` crashes right after this point, the replica node `B’`, which could be elected as the new primary,
+would not be aware of the slot ownership transfer without the synchronous replication of `SETSLOT`.
+This would leave the slot without an owner, leading to potential data loss and cluster topology inconsistency.
+
+#### Election in empty shards
+
 Starting from Valkey 8.0, Valkey clusters introduce the ability to elect a primary in empty shards.
 This behavior ensures that even when a shard is in the process of receiving its first slot,
 a primary can be elected. This prevents scenarios where there would be no primary available in the
 empty shard to handle redirected requests from the official slot owner,
 thereby maintaining availability during the live reconfiguration.
+
 ### ASK redirection
 
 In the previous section, we briefly talked about ASK redirection. Why can't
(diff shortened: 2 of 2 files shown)


topics/cluster-spec.md
```diff
@@ -479,6 +479,16 @@ is that:
...
 
+Starting from Valkey 8.0, the `CLUSTER SETSLOT` command is synchronously replicated to all healthy replicas
+running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
+If the replication fails, the primary does not execute the command, and the client receives a
+`NOREPLICAS Not enough good replicas to write` error. Operators can retry the command or customize the
+timeout using the `TIMEOUT` parameter to further increase the reliability of live reconfiguration:
+
+    CLUSTER SETSLOT slot [MIGRATING|IMPORTING|NODE] node-id [TIMEOUT timeout]
+
+Here, `timeout` is measured in seconds, with 0 meaning to wait indefinitely.
```
zuiderkwast, line 490: The timeout option should (also) be described in the command documentation: `commands/cluster-setslot.md`

topics/cluster-spec.md
```diff
@@ -513,6 +523,11 @@ set the slots to their normal state again. The same command is usually
 sent to all other nodes to avoid waiting for the natural
 propagation of the new configuration across the cluster.
 
+Starting from Valkey 8.0, Valkey clusters introduce the ability to elect a primary in empty shards.
+This behavior ensures that even when a shard is in the process of receiving its first slot,
+a primary can be elected. This prevents scenarios where there would be no primary available in the
+empty shard to handle redirected requests from the official slot owner,
+thereby maintaining availability during the live reconfiguration.
 ### ASK redirection
```
zuiderkwast, line 531: A blank line before the next heading never hurts and it's safer for correct rendering. (The rendering on the website could be different to how github renders it.)

topics/cluster-spec.md
```diff
@@ -550,6 +565,11 @@ Slots migration is explained in similar terms but with different wording
 (for the sake of redundancy in the documentation) in the `CLUSTER SETSLOT`
 command documentation.
 
+Starting from Valkey 8.0, when the primary in either the source or target shard fails during live reconfiguration,
+the primary in the other shard will automatically attempt to update its migrating/importing state to correctly pair
+with the newly elected primary. If this update is successful, the ASK redirection will continue functioning without
+requiring operator intervention.
```
zuiderkwast, line 571: Good.

But to complete the slot migration, the cluster does need manual intervention though, at least if valkey-cli is used, right? I mean if a failover happens while `valkey-cli cluster rehard` or `valkey-cli cluster rebalance` is running, valkey-cli aborts, right? Shouldn't we explain what an admin needs to do to complete the slot migration? `valkey-cli cluster --fix` can handle it?

Should we also mention that replicas are now able to return ASK redirects? I added it already to the READONLY command docs but maybe it's good to mention it briefly here in the ASK section in the cluster spec too?

This is what we have for the READONLY command regarding replicas sending redirects:

       3. Since Valkey 8.0, if a slot migration is ongoing.  In this case  the
          replica  can  return  an ASK redirect or a TRYAGAIN error reply.  In
          earlier versions, replicas are not aware of ongoing slot migrations.

topics/cluster-spec.md
```diff
@@ -479,6 +479,16 @@ is that:
 * All queries about existing keys are processed by "A".
 * All queries about non-existing keys in A are processed by "B", because "A" will redirect clients to "B".
 
+Starting from Valkey 8.0, the `CLUSTER SETSLOT` command is synchronously replicated to all healthy replicas
+running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
```
zuiderkwast, line 483: It's replicated only if all connected replicas are running valkey 8+, if I remember correctly.

```suggestion
Starting from Valkey 8.0, the `CLUSTER SETSLOT` command is synchronously replicated to all healthy replicas
if all replicas are running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
```

zuiderkwast (commented):
Awesome. But isn't it more? What I remember, there were a lot of changes and long code comments with scenario descriptions which are useful material for docs, like these:

* https://github.com/valkey-io/valkey/pull/245/files#diff-2515500619600c5a1e7a8d9aaa8761a6071314cd226c3b7ee4d0e054691f0a8eR3304
* https://github.com/valkey-io/valkey/pull/245/files#diff-2515500619600c5a1e7a8d9aaa8761a6071314cd226c3b7ee4d0e054691f0a8eR6330

Wouldn't it be good to copy these example scenarios to the relevant sections in the cluster spec?

topics/cluster-spec.md
PingXie: We are not requiring all replica nodes to be running 8.0. We went with eligible (in terms of both version and healthy state) nodes. I think rolling upgrades are a transient state. It should be acceptable to trade reliability for availability during this window, when upgrading from pre-8 to post 8 version.

PingXie:
> Awesome. But isn't it more? What I remember, there were a lot of changes and long code comments with scenario descriptions which are useful material for docs, like these:
> 
> * https://github.com/valkey-io/valkey/pull/245/files#diff-2515500619600c5a1e7a8d9aaa8761a6071314cd226c3b7ee4d0e054691f0a8eR3304
> * https://github.com/valkey-io/valkey/pull/245/files#diff-2515500619600c5a1e7a8d9aaa8761a6071314cd226c3b7ee4d0e054691f0a8eR6330
> 
> Wouldn't it be good to copy these example scenarios to the relevant sections in the cluster spec?

I wasn't sure how much detail we should add to the spec so went with just the "facts" but yeah good point on providing rationales as well. Will bring over some content.

btw, https://github.com/valkey-io/valkey/pull/245/files#diff-2515500619600c5a1e7a8d9aaa8761a6071314cd226c3b7ee4d0e054691f0a8eR3304 has been addressed by #754 and the comments have been removed (in #754 too).

topics/cluster-spec.md
zuiderkwast: OK, maybe I haven't fully understood this, but we use a shared replication buffer so all replicas receive the same stream of commands, right? In #686, we check for legacy replicas and if one is found, we don't replicate setslot to any replica. Correct me if I'm wrong.

zuiderkwast: I mean, we don't have to highlight this scenario, but the text should still be correct.

topics/cluster-spec.md
```diff
@@ -550,6 +581,16 @@ Slots migration is explained in similar terms but with different wording
...
 
+Starting from Valkey 8.0, when the primary in either the source or target shard fails during live reconfiguration,
+the primary in the other shard will automatically attempt to update its migrating/importing state to correctly pair
+with the newly elected primary. If this update is successful, the ASK redirection will continue functioning without
+requiring administrator intervention. In the event that slot migration fails, administrators can manually resume
+the interrupted slot migration by running the command `valkey-cli --cluster fix <ip:port>`.
+
+Additionally, since Valkey 8.0, replicas are now able to return `ASK` redirects during slot migrations.
+This capability was previously unavailable, as replicas were not aware of ongoing slot migrations in earlier versions.
+It's worth noting that this change has already been documented in the READONLY command section.
```
zuiderkwast, line 592: "already" and "worth noting", well, I added it a month ago, so roughly at the same time as this. We can just refer to READONLY if you want.

```suggestion
See the [READONLY](../commands/readonly.md) command.
```

topics/cluster-spec.md
zuiderkwast: Added text looks great. Thanks.

topics/cluster-spec.md
zuiderkwast: This is even explained in the PR description of #686:

> Older replicas do not support the CLUSTER SETSLOT command on replicas. If such a replica is found, the replication is skipped and falls back to the old (non-replicated) behavior.

commands/cluster-setslot.md
```diff
@@ -79,3 +79,12 @@ Notes:
   If the source node is informed before the destination node and the destination node crashes before it is set as new slot owner, the slot is left with no owner, even after a successful failover.
 * Step 6, sending `SETSLOT` to the nodes not involved in the resharding, is not technically necessary since the configuration will eventually propagate itself.
   However, it is a good idea to do so in order to stop nodes from pointing to the wrong node for the hash slot moved as soon as possible, resulting in less redirections to find the right node.
+* Starting from Valkey 8.0, `CLUSTER SETSLOT` is synchronously replicated to all healthy replicas
+  running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
+  If the replication fails, the primary does not execute the command, and the client receives a
+  `NOREPLICAS Not enough good replicas to write` error. Operators can retry the command or customize the
+  timeout using the `TIMEOUT` parameter to further increase the reliability of live reconfiguration:
+
+    CLUSTER SETSLOT slot [MIGRATING|IMPORTING|NODE] node-id [TIMEOUT timeout]
```
zuiderkwast, line 88: I can see this is not rendered as a code block. Probably because it's inside the bullet, so it needs two more spaces of indentation (the indentation of the bullet plus another 4 spaces for the code block) or, maybe better, use backticks:

````suggestion
  ```
  CLUSTER SETSLOT slot [MIGRATING|IMPORTING|NODE] node-id [TIMEOUT timeout]
  ```
````

File: topics/cluster-spec.md (hunk 4 above)
Comment (PingXie) [reply topics/cluster-spec.md]: Ah, sorry for the confusion. I see that I didn't word it accurately. I can fix that but I think the higher order question is what behavior do we like? "all or none" or "eligible replicas". I am in favor of "eligible replicas" because it is still better than no replication at all. Thoughts?

Comment (zuiderkwast) [reply topics/cluster-spec.md]: Replicating it only to "eligible replicas" would be better, but that's now what's currently implemented. That would require separate replication streams for legacy and new replicas. We have a single replication stream.

Can we just formulate this in a way that doesn't stress this difference? There's a difference only during rolling upgrade.

Comment (zuiderkwast) [reply topics/cluster-spec.md]: For example just omit the exact details. How about something like this?

> Starting from Valkey 8.0, the `CLUSTER SETSLOT` command is synchronously replicated if the replicas are running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.

File: commands/cluster-setslot.md [hunk 7]
```diff
@@ -79,3 +79,14 @@ Notes:
   If the source node is informed before the destination node and the destination node crashes before it is set as new slot owner, the slot is left with no owner, even after a successful failover.
 * Step 6, sending `SETSLOT` to the nodes not involved in the resharding, is not technically necessary since the configuration will eventually propagate itself.
   However, it is a good idea to do so in order to stop nodes from pointing to the wrong node for the hash slot moved as soon as possible, resulting in less redirections to find the right node.
+* Starting from Valkey 8.0, `CLUSTER SETSLOT` is synchronously replicated to all healthy replicas
```
madolson, line 82: I don't want to put synchronous replication anywhere in the docs. It's not technically synchronous since we don't revert it on failures, so it's only one copy of the data. I would prefer saying "is replicated to all healthy replicas before being executed on the primaries.

commands/cluster-setslot.md
```diff
@@ -79,3 +79,14 @@ Notes:
...
+  running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
+  If the replication fails, the primary does not execute the command, and the client receives a
+  `NOREPLICAS Not enough good replicas to write` error. Operators can retry the command or customize the
+  timeout using the `TIMEOUT` parameter to further increase the reliability of live reconfiguration:
```
madolson, line 86: Btw, we should either decide if this style guide is wrong or not: https://github.com/valkey-io/valkey-doc?tab=readme-ov-file#styling-guidelines. "Start every sentence on a new line."

We didn't force it in the past to retain history, but I find it annoying sometimes follow it.

topics/cluster-spec.md
```diff
@@ -513,6 +513,37 @@ set the slots to their normal state again. The same command is usually
...
+Synchronous replication is critical for maintaining cluster consistency during live reconfiguration.
+Before applying changes like slot ownership and migrating states to the primary, these must be fully
+replicated to all replicas. This prevents loss of state if the primary fails after executing the command.
+
+Consider a scenario where the target primary node `B` is finalizing a slot migration.
+Before the `SETSLOT` command is replicated to its replica node `B’`, `B` might send a cluster `PONG`
+message to the source primary node `A`, promoting `A` to relinquish its ownership of the slot in question.
+If `B` crashes right after this point, the replica node `B’`, which could be elected as the new primary,
+would not be aware of the slot ownership transfer without the synchronous replication of `SETSLOT`.
+This would leave the slot without an owner, leading to potential data loss and cluster topology inconsistency.
```
madolson, line 537:
```suggestion
Replicating the slot information prevents loss of state if the primary fails after executing the command.
Consider a scenario where the target primary node `B` is finalizing a slot migration.
Before the `SETSLOT` command is replicated to its replica node `B’`, `B` might send a cluster `PONG`
message to the source primary node `A`, promoting `A` to relinquish its ownership of the slot in question.
If `B` crashes right after this point, the replica node `B’`, which could be elected as the new primary,
would not be aware of the slot ownership transfer without the synchronous replication of `SETSLOT`.
This would leave the slot without an owner, leading to potential data loss and cluster topology inconsistency.
```

topics/cluster-spec.md
```diff
@@ -513,6 +513,37 @@ set the slots to their normal state again. The same command is usually
 sent to all other nodes to avoid waiting for the natural
 propagation of the new configuration across the cluster.
 
+#### Synchronous replication of `CLUSTER SETSLOT`
+
+Starting from Valkey 8.0, the `CLUSTER SETSLOT` command is synchronously replicated to all healthy replicas
+running Valkey version 8.0+. By default, this synchronous replication must complete within 2 seconds.
+If the replication fails, the primary does not execute the command, and the client receives a
+`NOREPLICAS Not enough good replicas to write` error. Operators can retry the command or customize the
+timeout using the `TIMEOUT` parameter to further increase the reliability of live reconfiguration:
```
madolson, line 522:
```suggestion
timeout using the `TIMEOUT` parameter to further increase the reliability of live resharding:
```
Everywhere else in this doc we call it live resharding and not "live reconfiguration". I think we should be consistent.

madolson (commented):
To viktors comment, I would prefer less detail in the cluster-spec when not explicitly needed by the end user. A lot of the "fixes" are already basically hidden when things go correctly. I think the current level of detail seems right to me.

commands/cluster-setslot.md
zuiderkwast: I don't mind dropping it. I don't like to complain about it, so I haven't enforced it.

We can just skip reflowing text when it's edited.

commands/cluster-setslot.md
PingXie: >since we don't revert it on failures, so it's only one copy of the data.

I don't think this is the common definition of  "synchronous" replication. I modeled our "synchronous" replication after [postgress'](since we don't revert it on failures, so it's only one copy of the data.) 

> I would prefer saying "is replicated to all healthy replicas before being executed on the primaries.

This is also not accurate nor complete because we do wait for a certain number of replicas to ack the replication and this is the core part of this feature offering.

commands/cluster-setslot.md
PingXie: yeah can we consider dropping this rule? I don't see this being enforced today on the existing doc. I feel that there are a few main situations wherein we need to modify a doc

1. either we fix typos (or small changes)
2. or we rewrite some portion of the doc because we have changed the behavior 
3. or we add new content like in this case

In both the second and third cases, I would think we will have to review the entire paragraph already. I don't think we should ever re-flow the first case, irrespective of how we wrote the doc in the first place.

thoughts?

topics/cluster-spec.md
PingXie: sounds good to me

topics/cluster-spec.md
PingXie: I picked this phrase (live reconfiguration) from the section title https://github.com/valkey-io/valkey-doc/pull/167/files#diff-93c993e7eca8dc9073483f2c2bc33f13fd9ebd7c2a3a3657b364bca84d5f289eL406 :)

This is also pretty much the first time I came across this terminology. Any concern if I update the section title to "live resharding" too for consistency?

commands/cluster-setslot.md
zuiderkwast: I guess "reconfiguration" includes adding and removing nodes. Resharding is only about moving slots?

I'm fine with either of the terms. :)

topics/cluster-spec.md
zuiderkwast: I guess "reconfiguration" includes adding and removing nodes. Resharding is only about moving slots?

I'm fine with either of the terms. :)

PingXie: Yes, but these concepts are intertwined. We mainly move slots due to node arrivals or departures. :)  While I slightly prefer 'resharding' at the moment, 'reconfiguration' also works for me. I do feel strongly though about consistency.

commands/cluster-setslot.md
madolson: > I don't think this is the common definition of "synchronous" replication. I modeled our "synchronous" replication after [postgress'](https://www.postgresql.org/docs/current/warm-standby.html#SYNCHRONOUS-REPLICATION). The "revert" part falls in the realm of "distributed transactions" and it is a step up from "synchronous replication".

In postgreSQL, everything is a transaction though. There is no "step up" so to speak, they are the same.

> This is also not accurate nor complete because we do wait for a certain number of replicas to ack the replication and this is the core part of this feature offering.

"Is replicated and acknlowedged on all healthy replicas before being executed on the primary"?

PingXie: >In postgreSQL, everything is a transaction though.

I am not aware that Postgress' replication is done transactionally based on my research. My understanding is that there is no rollback on other replicas when one replica fails to ack the replication either. It is just that the primary will wait for replicas to ack the replication. Note that in order to achieve transactional replication, one would need to implement 2PC, which trades (more) performance/availability for consistency. Do you have a Postgress code or doc pointer that indicates their repliction is done transactionally?  I can dig some more too.

>"Is replicated and acknlowedged on all healthy replicas before being executed on the primary"?

Yes. 

The "synchronous" qualifier is critical here and it is used to contrast the current "asynchronous" replication, which doesn't provide the required semantics. 

BTW, not related to Postgress, but "synchronous replication" is also used in the exact same semantics as I used here for another database I worked on before.

zuiderkwast: I vote for explaining this without using the word "synchronous". It's controversial to introduce this term.

PingXie: >However, we don't guarantee synchronous replication since the replicas can be down at the time, which is my bigger concern at the moment

I am convinced by your argument that this is not a strict synchronous replication so "sync repl" is indeed overselling. However, I still think we need a name for this concept to differentiate from the normal replication behavior today.

I think the concept we implemented is closer to MySQL's [semi-synchronous replication](https://dev.mysql.com/doc/refman/8.4/en/replication-semisync.html) but even that offers more guarantee. I wonder if "Opportunistic Synchronous Replication" would capture our idea more accurately?

>Semisynchronous replication falls between asynchronous and fully synchronous replication. The source waits until at least one replica has received and logged the events (the required number of replicas is configurable), and then commits the transaction. The source does not wait for all replicas to acknowledge receipt, and it requires only an acknowledgement from the replicas, not that the events have been fully executed and committed on the replica side. Semisynchronous replication therefore guarantees that if the source crashes, all the transactions that it has committed have been transmitted to at least one replica.

zuiderkwast: I'm not sure anymore. We actually do wait for "all healthy replicas" before committing, so this seems synchronous to me. If we timeout, we don't rollback this from the replicas that already ACKed this though.

madolson: I'd still vote for no name. I asked a bunch of folks at AWS. The most common response was "don't invent a name" if it's not an industry standard, since that just adds more confusion to this mix. "Best effort synchronous replication" came up from a few folks, which I don't really like. Someone suggested "replicated with acknowledgement", which I kind of like but not sure if it resonates as a "good name".

PingXie: "Best effort" == "Opportunistic" :) I think we will need a name at some point IF we expand this pattern beyond this command but I am on board with not adding a new concept for this one. I will drop "synchronous" but keep the high level explanation in this doc.

zuiderkwast: Best effort is kind of the key concept since the very beginning. Memcached doesn't have persistence and replication. Redis/Valkey has it, best effort, async, without compromising latency.

The cluster bus is supposed to be truly consistent though. We'll fix that some day.

zuiderkwast: Btw, the MIGRATE command doesn't use the replication nor cluster bus. It just sets up a new connection. Is it synchronous? It waits for OK before deleting the transferred key, I think, or does it? Is this how we should have done SETSLOT?

madolson: > "Best effort" == "Opportunistic"

Yes. I would be OK with either of those two FWIW, it's not my top preference.

PingXie: I dropped "synchronous". PTAL

PingXie:
@madolson @zuiderkwast any more feedback? Let's merge this one in the next day or two?

zuiderkwast (approved):
This looks good now. Thanks! (Nothing controversial. :laughing:)

madolson (approved):
Ship it!

PR #179, opened 2024-10-09
No persistence

melroy89:
Explain how to configure no persistence (eg. when using Valkey as a cache, without the need for storage).

Related discussion: https://github.com/orgs/valkey-io/discussions/1134

### DIFF
--- a/topics/persistence.md
+++ b/topics/persistence.md
@@ -77,6 +77,8 @@ Then add the following configuration instead:
 
     save ""
 
+Alternatively, you can also use the `--save ""` argument on the `valkey-server` binary.
+
 ### How it works
 
 Whenever Valkey needs to dump the dataset to disk, this is what happens:
--- a/topics/persistence.md
+++ b/topics/persistence.md
@@ -70,7 +70,7 @@ This strategy is known as _snapshotting_.
 
 ## No persistence
 
-If you prefer **not** to have persistence (eg. when using a Valkey instance solely as a cache) that is also a possibility.
+If you prefer **not** to have persistence (for example when using a Valkey instance solely as a cache) that is also a possibility.
 
 RDB snapshotting is enabled by default. To disable it, set the `save` configuration parameter to the empty string `""` and remove any excess `save` lines that are present in the current configuration file.
 
--- a/topics/persistence.md
+++ b/topics/persistence.md
@@ -68,6 +68,15 @@ dataset to disk every 60 seconds if at least 1000 keys changed:
 
 This strategy is known as _snapshotting_.
 
+## No persistence
+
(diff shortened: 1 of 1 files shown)


amirreza8002:
hi
maybe the command line options should be mentioned too.
for my case (docker) command line seems much easier

melroy89:
> hi maybe the command line options should be mentioned too. for my case (docker) command line seems much easier

Added!

topics/persistence.md
```diff
@@ -68,6 +68,17 @@ dataset to disk every 60 seconds if at least 1000 keys changed:
 
 This strategy is known as _snapshotting_.
 
+## No persistence
+
+If you prefer **not** to have persistence (eg. when using a Valkey instance solely as a cache) that is also possibility.
```
hpatro, line 73: *a possibility

topics/persistence.md
```diff
@@ -68,6 +68,17 @@ dataset to disk every 60 seconds if at least 1000 keys changed:
 
 This strategy is known as _snapshotting_.
 
+## No persistence
+
+If you prefer **not** to have persistence (eg. when using a Valkey instance solely as a cache) that is also a possibility.
+
+First comment out or remove all existing `save` lines that are present in the current configuration file.   
+Then add the following configuration instead:
```
zuiderkwast, line 76: This sort of assumes there's a lot of junk in the user's config file. I'd prefer to focus first on the correct config and only afterwards mention what to do with extra junk in the files, if any. How about this?

```suggestion
RDB snapshotting is enabled by default. To disable it, set the `save` configuration parameter to the empty string `""` and remove any excess `save` lines that are present in the current configuration file.
```

zuiderkwast (commented):
Very nice, thanks! I added a small suggestion.

topics/persistence.md
```diff
@@ -68,6 +68,17 @@ dataset to disk every 60 seconds if at least 1000 keys changed:
 
 This strategy is known as _snapshotting_.
 
+## No persistence
+
+If you prefer **not** to have persistence (eg. when using a Valkey instance solely as a cache) that is also a possibility.
```
zuiderkwast, line 73: The spellchecker complains about "eg." I think it wants it spelled "e.g.", but I recall now that we've avoided abbreviatinos like this in the docs in the past. Maybe "for example" is clearer?

```suggestion
If you prefer **not** to have persistence (for example when using a Valkey instance solely as a cache) that is also a possibility.
```

melroy89: Agreed my bad. I also also reluctant.

zuiderkwast (approved):
Aweseome. Thanks!

If you want help more with docs, there are some open issues in this repo. :)

PR #184, opened 2024-10-25
Corrections to migration guide

zuiderkwast:
- User the term "Redis OSS" and "open source Redis", rather than "Redis Community Edition". The latter refers to the source-available versions from 7.4 and later.
- Mention that Valkey is backward compatible with all old Redis versions.
- Update the migration table to include Redis OSS 2.x - 7.2.x. Users have been confused by the table that seems to indicate that it's not possible to upgrade from Redis 6.2 to Valkey 8.0, but it is.
- Mention that it may be possible to migrate from proprietary Redis, but that it's not covered by this document. (For example, writing the data to disk using AOF without preamble and loading in Valkey can probably work.)

### DIFF
--- a/topics/migration.md
+++ b/topics/migration.md
@@ -4,10 +4,10 @@ linkTitle: Migration
 description: How to migrate from Redis to Valkey
 ---
 
-This is a migration guide from Redis Community Edition to Valkey. 
+This is a migration guide from Redis open source versions to Valkey.
 You will learn how to migrate a standalone Redis server instance and a Redis Cluster. 
 
-This guide provides migration steps for Redis server and Valkey deployed in Docker; however, they should also apply for on-premises deployments. 
+This guide provides migration steps for Redis server and Valkey deployed in Docker; however, they should also apply for other deployments.
 Refer to [install Valkey](installation.md) for installation options.
 
 ## Why to migrate to Valkey?
@@ -18,15 +18,19 @@ Refer to [install Valkey](installation.md) for installation options.
 
 ### Migration compatibility matrix
 
-You can migrate Redis server to Valkey. 
-The following table provides migration options depending on the Redis version you run:
+You can migrate a Redis server to Valkey.
+Valkey is backward compatible with Redis OSS 7.2 and all earlier open source Redis versions.
+Migrating from any open source Redis version to Valkey is just an upgrade.
+Redis 7.4 and later are not open source and the data files are no longer compatible with Valkey.
+It may be possible to migrate the data to Valkey from proprietary Redis versions and other Redis-like software but it requires another method and is not covered by this document.
 
-| Redis | Valkey |
-|-------|--------|
-| 6.x   | 7.2.x  |
-| 7.2.x | 7.2.x  |
-| 7.2.x | 8.0.x  |
-| 7.4   | n/a    |
+The following table provides migration options depending on the open source Redis version you run:
+
+| Redis                 | Valkey |
+|-----------------------|--------|
+| 2.x - 7.2.x           | 7.2.x  |
+| 2.x - 7.2.x           | 8.0    |
+| 7.4 (not open source) | n/a    |
 
 ## Migrate a standalone instance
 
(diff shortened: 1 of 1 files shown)


nastena1606 (approved):
LGTM

topics/migration.md
```diff
@@ -18,15 +18,19 @@ Refer to [install Valkey](installation.md) for installation options.
 
 ### Migration compatibility matrix
 
-You can migrate Redis server to Valkey. 
-The following table provides migration options depending on the Redis version you run:
+You can migrate a Redis server to Valkey.
+Valkey is backward compatible with Redis OSS 7.2 and all earlier open source Redis versions.
```
madolson, line 22:
```suggestion
Valkey is compatible with Redis OSS 7.2 and all earlier open source Redis versions.
```
It's a different engine, it's not backwards compatibility.

topics/migration.md
```diff
@@ -18,15 +18,19 @@ Refer to [install Valkey](installation.md) for installation options.
...
+Migrating from any open source Redis version to Valkey is just an upgrade.
+Redis 7.4 and later are not open source and the data files are no longer compatible with Valkey.
```
madolson, line 24:
```suggestion
Redis Community Edition (CE), versions 7.4 and later, are not open source and the data files are not compatible with Valkey.
```

topics/migration.md
```diff
@@ -18,15 +18,19 @@ Refer to [install Valkey](installation.md) for installation options.
...
-| 7.2.x | 7.2.x  |
-| 7.2.x | 8.0.x  |
-| 7.4   | n/a    |
+The following table provides migration options depending on the open source Redis version you run:
+
+| Redis                 | Valkey |
+|-----------------------|--------|
+| 2.x - 7.2.x           | 7.2.x  |
+| 2.x - 7.2.x           | 8.0    |
+| 7.4 (not open source) | n/a    |
```
madolson, line 33:
```suggestion
The following table provides migration options depending on the Redis version you run:

| Redis                 | Valkey |
|-----------------------|--------|
| OSS 2.x - 7.2.x       | 7.2.x  |
| OSS 2.x - 7.2.x       | 8.0    |
| CE 7.4                | n/a    |
```

I found it odd that you added open source in the header, but then later called it non-OSS. Maybe this is clearer?

topics/migration.md
```diff
@@ -18,15 +18,19 @@ Refer to [install Valkey](installation.md) for installation options.
 
 ### Migration compatibility matrix
 
-You can migrate Redis server to Valkey. 
+You can migrate a Redis server to Valkey.
+Valkey is compatible with Redis OSS 7.2 and all earlier open source Redis versions.
+Migrating from any open source Redis version to Valkey is just an upgrade.
```
stockholmux, line 23: I would change 'just' -> 'effectively'

topics/migration.md
```diff
@@ -18,15 +18,19 @@ Refer to [install Valkey](installation.md) for installation options.
...
+Redis Community Edition (CE), versions 7.4 and later, are not open source and the data files are not compatible with Valkey.
```
stockholmux, line 24: I assume you mean RDB files? I don't think we use the term 'data files' often. If there is a link that explains the difference between 7.4 and Valkey that would be helpf to the reader.

stockholmux (changes requested):

topics/migration.md
madolson: It would also include AOF files I presume, so not just RDB.

madolson (approved):
I'm OK with it as is, but think Kyle's suggestions also make sense.

topics/migration.md
zuiderkwast:
```suggestion
Migrating from any open source Redis version to Valkey is effectively an upgrade.
```

topics/migration.md
zuiderkwast: I don't have time to go into the details how you can migrate. It's possible to use various tools, like rdblib and RediShake, that can convert between different RDB versions, connect as a replica and emit commands to a Valkey node, or something like that. To do this without any service impact is probably possible, but I don't have enough experience with it.

I just want to remove the parts that are incorrect in this document, and to make it not imply that it's impossible to migrate between any other versions than those listed in the table.

zuiderkwast: Btw, Kyle, I agree with your comments. It's just a matter of prioritization.

PR #189, opened 2024-11-20
Documentation for SET IFEQ

sarthakaggarwal97:
This PR adds the documentation to support https://github.com/valkey-io/valkey/pull/1324

### DIFF
--- a/commands/set.md
+++ b/commands/set.md
@@ -12,22 +12,30 @@ The `SET` command supports a set of options that modify its behavior:
 * `PXAT` *timestamp-milliseconds* -- Set the specified Unix time at which the key will expire, in milliseconds (a positive integer).
 * `NX` -- Only set the key if it does not already exist.
 * `XX` -- Only set the key if it already exists.
+* `IFEQ` -- Set the key if the given value matches the existing value. An error is returned and `SET` aborted if the value stored at key is not a string.
 * `KEEPTTL` -- Retain the time to live associated with the key.
 * `GET` -- Return the old string stored at key, or nil if key did not exist. An error is returned and `SET` aborted if the value stored at key is not a string.
-* `IFEQ` -- Set the key if the given value matches the existing value. An error is returned and `SET` aborted if the value stored at key is not a string.
 
 Note: Since the `SET` command options can replace `SETNX`, `SETEX`, `PSETEX`, `GETSET`, it is possible that in future versions of Valkey these commands will be deprecated and finally removed.
 
 ## Examples
 
+Basic usage
 ```
 127.0.0.1:6379> SET mykey "Hello"
 OK
 127.0.0.1:6379> GET mykey
 "Hello"
+```
+
+Set a value and an expiry time.
+```
 127.0.0.1:6379> SET anotherkey "will expire in a minute" EX 60
 OK
+```
 
+Conditionally set a value.
+```
 127.0.0.1:6379> SET foo "Initial Value"
 OK
 127.0.0.1:6379> GET foo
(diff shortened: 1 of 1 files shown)


sarthakaggarwal97:
@hpatro raised the doc PR for IFEQ.

commands/set.md
```diff
@@ -13,7 +13,8 @@ The `SET` command supports a set of options that modify its behavior:
 * `NX` -- Only set the key if it does not already exist.
 * `XX` -- Only set the key if it already exists.
 * `KEEPTTL` -- Retain the time to live associated with the key.
-* `!GET` -- Return the old string stored at key, or nil if key did not exist. An error is returned and `SET` aborted if the value stored at key is not a string.
+* `GET` -- Return the old string stored at key, or nil if key did not exist. An error is returned and `SET` aborted if the value stored at key is not a string.
+* `IFEQ` -- Set the key if the given value matches the existing value. An error is returned and `SET` aborted if the value stored at key is not a string.
```
zuiderkwast, line 17: Let's put IFEQ after NX and XX, because they are related.

Remove "An error is returned and `SET` aborted if the value stored at key is not a string.". It's not specific to the IFEQ option, is it?

commands/set.md
```diff
@@ -24,9 +25,17 @@ Note: Since the `SET` command options can replace `SETNX`, `SETEX`, `PSETEX`, `G
 OK
 127.0.0.1:6379> GET mykey
 "Hello"
-127.0.0.1:6379> 
 127.0.0.1:6379> SET anotherkey "will expire in a minute" EX 60
 OK
+
+127.0.0.1:6379> SET foo "Initial Value"
```
zuiderkwast, line 31: The blank line will never be printed in the valkey-cli session.

Either...

* remove this blank line so all examples are in the same valkey-cli session,
* or split the examples into separate valkey-cli session and put a short text for each example.

If you decide to split the example, I mean something like this:

> Basic usage.
>
> ```
> 127.0.0.1:6379> SET mykey "Hello"
> OK
> 127.0.0.1:6379> GET mykey
> "Hello"
> ```
>
> Set a value and an expiry time.
>
> ```
> 127.0.0.1:6379> SET anotherkey "will expire in a minute" EX 60
> OK
> ```
>
> Conditionally set a value.
>
> ```
> 127.0.0.1:6379> SET foo "Initial Value"
> ...
> ```

zuiderkwast (commented):
Thanks. Just a few nits.

commands/set.md
zuiderkwast: Sorry! I see now that "An error is returned and `SET` aborted if the value stored at key is not a string." is special code for IFEQ. Let's keep this.

sarthakaggarwal97: addressed @zuiderkwast

commands/set.md
sarthakaggarwal97: I have split them, thank you!

commands/set.md
```diff
@@ -12,23 +12,40 @@ The `SET` command supports a set of options that modify its behavior:
 * `PXAT` *timestamp-milliseconds* -- Set the specified Unix time at which the key will expire, in milliseconds (a positive integer).
 * `NX` -- Only set the key if it does not already exist.
 * `XX` -- Only set the key if it already exists.
+* `IFEQ` -- Set the key if the given value matches the existing value. An error is returned and `SET` aborted if the value stored at key is not a string.
```
zuiderkwast, line 15: We should include the argument too, like the other ones with arguments above (EX, PX, etc.).

```suggestion
* `IFEQ` *comparison-value* -- Set the key if the comparison value matches the existing value. An error is returned and `SET` aborted if the value stored at key is not a string.
```

zuiderkwast (approved):
Great! Thanks

sarthakaggarwal97:
Thank you @zuiderkwast :D

PR #192, opened 2024-11-26
Add ValkeyModule_ReplicateWithFlag description

sungming2:
Follow up documentation for new module API(ValkeyModule_ReplicateWithFlag) https://github.com/valkey-io/valkey/pull/1357


zuiderkwast:
Actually we don't need a doc PR for module API.

The API docs are extracted from comments in module.c using a script under utils/ in the valkey repo. I will do it before we release the next minor version.

It means that the docs are written in module.c so just make sure those are correct.

Closing this. Thanks.

PR #193, opened 2024-11-28
No valkey in titles

zuiderkwast:
* Deletes "Valkey" in titles, because "Valkey" or "Valkey documentation" is added to each page title when the page is rendered.
* Deletes unused `linkTitle` frontmatter field.


stockholmux (approved):
LGTM

PR #195, opened 2024-12-07
Update PHP Redlock impl. doc

mvorisek:
Fix the PHP Redlock impl. link and move it as the 1st mentioned PHP impl. as it is the most popular and the only library which supports both https://github.com/phpredis/phpredis and https://github.com/predis/predis clients.

PR for Redis docs: https://github.com/redis/docs/pull/964


zuiderkwast (approved):
OK, thanks! I don't know these, but I can see this one is updated and the other one is very old. :)

PR #199, opened 2024-12-12
Explanation of issues when using GET and NX/XX/IFEQ together in SET command

soloestoy:


commands/set.md
```diff
@@ -15,6 +15,7 @@ The `SET` command supports a set of options that modify its behavior:
 * `IFEQ` *comparison-value* -- Set the key if the comparison value matches the existing value. An error is returned and `SET` aborted if the value stored at key is not a string.
 * `KEEPTTL` -- Retain the time to live associated with the key.
 * `GET` -- Return the old string stored at key, or nil if key did not exist. An error is returned and `SET` aborted if the value stored at key is not a string.
+    * Note: When `GET` is used together with `NX`/`XX`/`IFEQ`, it is impossible to determine whether the `SET` command executed successfully based on the reply.
```
zuiderkwast, line 18: It's impossible based on the reply alone, but it's not impossible based on the reply combined with some other knowledge the client has.

```suggestion
    * Note: When `GET` is used together with `NX`/`XX`/`IFEQ`, it is impossible to determine whether the `SET` command executed successfully based on the reply alone.
```

zuiderkwast (commented):
Great. I have a very small suggestion.

zuiderkwast:
I think it's a small disaster that the Reply section is missing on the website version of the docs.

The replies are documented in the files `resp2_replies.json` and `resp3_replies.json`. Assuming it will be included on the website some time in the future, we can consider improving this part too.

This is the man page version of the reply section:

```
REPLY
   RESP2
       Any of the following:

       • valkey-protocol(7) Nil reply: GET not given: Operation was aborted (conflict with one of the XX/NX options).

       • valkey-protocol(7) Simple string reply: OK.  GET not given: The key was set.

       • valkey-protocol(7) Nil reply: GET given: The key didn’t exist before the SET.

       • valkey-protocol(7) Bulk string reply: GET given: The previous value of the key.

   RESP3
       Any of the following:

       • valkey-protocol(7) Null reply: GET not given: Operation was aborted (conflict with one of the XX/NX options).

       • valkey-protocol(7) Simple string reply: OK.  GET not given: The key was set.

       • valkey-protocol(7) Null reply: GET given: The key didn’t exist before the SET.

       • valkey-protocol(7) Bulk string reply: GET given: The previous value of the key.
```

Issue #154, opened 2024-07-04
Document configs


zuiderkwast:
Currently, we just refer to valkey.conf (the template config file) where the fields are described in comments. This isn't very good documentation.

Document each config parameter in the proper page [valkey.conf.md](https://github.com/valkey-io/valkey-doc/tree/main/topics/valkey.conf.md). Instead of referring to the config files of different versions, we can mention for each field in which version it was introduced.

rjd15372:
I also think this is a good idea. Having the config options description in our documentation is the way to go. But there's a maintenance burden here that we need to decide how to tackle on, because we will have duplicate information (in the valkey.conf file, and here in the documentation).

My take on this is that whenever a developer opens a PR with changes to the `valkey.conf` file in the valkey repo, the reviewers should ask the developer to also open a PR in the valkey-doc repo with updating the documentation. This way we distribute the burden of maintenance between all developers.

zuiderkwast:
Yes, for any feature that requires doc changes, we ask the developer to do it and we add the tag 'doc-pr-needed' to the PR.

Can I assign you to this issue? ;)

rjd15372:
@zuiderkwast you can assign it to @nastena1606

zuiderkwast:
Ideas:

1. Minimal. Change the link to `valkey.conf` to point to valkey.conf in unstable branch. Remove the old links to Valkey 7 config files and old redis.conf.
2. Automated. We could write a script to include the contents of `valkey.conf` (the example config with comments) in a code block on this page. In this way, contributors that modify configs don't need to make a doc PR. We only need to run the script when we make a new release.
3. Full. Add proper documentation of the configs in this file. Some drawbacks: We will have duplicated information (new docs + the comments valkey.conf) and contributors will have to make a doc PR when changing some configs.

melroy89:
> What we talked about in weekly core meeting:
> 
>     1. We want to pull out the documentation from the valkey.conf file. Where and how is not exactly defined yet, but we would like to have it stored in markdown/json/yaml/toml somewhere else.
> 
>     2. In Valkey 10 we will stop updating the valkey.conf file. We will find a way to keep them in sync with each other, the valkey.conf file will not be authoritative.

sounds like a good plan.

PR #205, opened 2024-12-27
Introduce import-mode and client import-source

lyq2333:
This PR is for https://github.com/valkey-io/valkey/pull/1185 and https://github.com/valkey-io/valkey/pull/1398.

By far, RedisShake doesn't support `import-mode` and `client import-source` yet. So I don't add a new migration guide in migration.md. If any sync tools support import-mode, I'm willing to supplement it.


commands/client-import-source.md
```diff
@@ -0,0 +1,6 @@
+When the client sync data from the Redis-Like server to Valkey using some sync tools like [RedisShake](https://github.com/tair-opensource/RedisShake), Valkey performs expiration and eviction as usual because it's a primary, which may cause data corruption. For example, the client calls `set key 1 ex 1` on the source server and this command is transferred to the destination server. Then the client calls `incr key` on the source server before the key expired, there will be a key on the source server with the value of 2. But when the command arrived at the destination server, the key may be expired and has deleted. So there will be a key on the destination server with the value of 1, which is inconsistent with the source server. Sync tools can solve this problem by setting ` IMPORT-MODE` config to yes and declaring their connections as  `IMPORT-SOURCE`.
```
zuiderkwast, line 1:
```suggestion
When the client sync data from a Redis-like server to Valkey using some sync tools like [RedisShake](https://github.com/tair-opensource/RedisShake), Valkey performs expiration and eviction as usual because it's a primary, which may cause data corruption. For example, the client calls `set key 1 ex 1` on the source server and this command is transferred to the destination server. Then the client calls `incr key` on the source server before the key expired, there will be a key on the source server with the value of 2. But when the command arrived at the destination server, the key may be expired and has deleted. So there will be a key on the destination server with the value of 1, which is inconsistent with the source server. Sync tools can solve this problem by setting `import-mode` config to `yes` and declaring their connections as `IMPORT-SOURCE`.
```

commands/client-import-source.md
```diff
@@ -0,0 +1,6 @@
...
+
+The `CLIENT IMPORT-SOURCE` command mark this client as an import source if import-mode is enabled and can visit expired keys. The following modes are available:
+
+* `ON`. In this mode the client can visit expired keys.
+* `OFF`. This is the default mode in which the client works as a normal client.
```
zuiderkwast, line 6: I think the config is very important. We can even mention it in a separate Notes section. (Notes is a standard section name in man pages.)

```suggestion
* `OFF`. This is the default mode in which the client works as a normal client.

## Notes

The server needs to be configured with `import-mode yes` before marking a client connection as an import source.
```

zuiderkwast (approved):
Thanks!

Just some nits. I can fix them before merging.

Still on my wish list is a real documentation page for all configs, #154. I hope someone will fix it, some time...

PR #206, opened 2024-12-30
SET: Be more explicit about replies when using GET in conjunction with XX/NX/IFEQ

gmbnomis:
Based on the conversation/comment https://github.com/valkey-io/valkey/issues/1415#issuecomment-2563154267

With PR https://github.com/valkey-io/valkey-io.github.io/pull/183, the "RESP2 Reply" section looks like this for the web page:

image


zuiderkwast (approved):
Thanks, this is very clear.

PR #207, opened 2025-01-07
Add availability_zone to the HELLO response.

rueian:
This PR is for https://github.com/valkey-io/valkey/pull/1487.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -6831,6 +6831,10 @@
         "complexity": "O(1)",
         "history": [
             [
+                "8.1.0",
+                "A new `availability_zone` field has been added into the response if the `availability_zone` config is set."
+            ],
+            [
                 "6.2.0",
                 "`protover` made optional; when called without arguments the command reports the current connection's context."
             ]


commands.json
```diff
@@ -6830,6 +6830,10 @@
         "group": "connection",
         "complexity": "O(1)",
         "history": [
+            [
+                "8.1.0",
+                "A new `availability_zone` field has been added into the response if the `availability_zone` config is set."
```
zuiderkwast, line 6835: This file `commands.json` was probably generated from valkey (or redis) using one of the scripts under utils/ in the valkey repo. I don't know exactly how it is done. (Probably we should delete it, because I think it is not up to date.) We don't use this file for anything in the website or man pages. We use the information directly from Valkey instead.

What to update instead is the HELLO page: https://github.com/valkey-io/valkey-doc/blob/main/commands/hello.md

rueian: Oh, I see. But I think the content of the hello.md is fine and I'd like to update the History part instead:
image

So I guess what I should do is go back to the main repo and update the https://github.com/valkey-io/valkey/blob/3b52186b6a6199884edf447c52828d5ca882142f/src/commands/hello.json

rueian: The updated PR is https://github.com/valkey-io/valkey/pull/1524.

madolson:
Don't we still need to update this file? https://github.com/valkey-io/valkey-doc/blob/main/commands/hello.md

zuiderkwast:
`hello.md` mentions all fields just in an example, but at least they are visible.

I think we should add a sentence that mentions that the fields in the example are always present and the availability_zone field is present only if AZ is configured.

rueian:
Hi @zuiderkwast and @madolson,

The update for hello.md is in the new PR https://github.com/valkey-io/valkey-doc/pull/208.

Please let me know if it looks good or not.

PR #211, opened 2025-01-12
RESP2/RESP3 reply unification and minor fixes

gmbnomis:
This allows to merge more of the RESP2/RESP3 command replies.


zuiderkwast (approved):
This is great, thanks!

PR #217, opened 2025-01-15
Add details for testing RDMA server

z-koki:
Add details about RDMA server functionality test.

### DIFF
--- a/topics/RDMA.md
+++ b/topics/RDMA.md
@@ -40,6 +40,8 @@ Or append `loadmodule src/valkey-rdma.so bind=192.168.122.100 port=6379` in valk
 
     ./src/valkey-server valkey.conf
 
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
Thanks!

There is a plan that valkey-cli will support RDMA in 9.0 but not yet. We will need to update this information again...

@pizhenwei FYI

pizhenwei (commented):
I'm confused about this, `valkey-cli` does not support RDMA now because of the RDMA support from libhiredis, there is no relationship with `IPoIB`.

z-koki:
> I'm confused about this, `valkey-cli` does not support RDMA now because of the lack of RDMA support from libhiredis, there is no relationship with `IPoIB`.

很抱歉这么久才回复。我当时测试的环境是IB网络，因此如果服务器使用`valkey-server bind=$ipoib_addr`参数运行，客户端指定`$ipoib_addr`，客户端也能跑起来，但是实际上是跑在IPoIB上，而不是真正的RDMA。 #264 的措辞更合适，非常感谢！

I apologize for the delayed response. The environment I was testing in at the time was an IB network. Therefore, if the server was run with the parameter `valkey-server bind=$ipoib_addr` and the client specified `$ipoib_addr`, the client could also run, but it was actually running over IPoIB, not true RDMA. The wording in #264  is more appropriate, thank you very much!

PR #218, opened 2025-01-16
Updated Documentation for CLIENT KILL and CLIENT LIST

sarthakaggarwal97:
Documentation for https://github.com/valkey-io/valkey/pull/1401

### DIFF
--- a/commands/client-kill.md
+++ b/commands/client-kill.md
@@ -13,7 +13,7 @@ instead of killing just by address. The following filters are available:
 
 * `CLIENT KILL ADDR ip:port`. This is exactly the same as the old three-arguments behavior.
 * `CLIENT KILL LADDR ip:port`. Kill all clients connected to specified local (bind) address.
-* `CLIENT KILL ID client-id [client-id ...]`. Allows to kill a client by its unique `ID` field. Client `ID`'s can be retrieved using the `CLIENT LIST` command. The filter supports one or more `client-id` arguments
+* `CLIENT KILL ID client-id [client-id ...]`. Allows to kill a client by its unique `ID` field. Client `ID`'s can be retrieved using the `CLIENT LIST` command. The filter supports one or more `client-id` arguments.
 * `CLIENT KILL TYPE type`, where *type* is one of `normal`, `master`, `replica` and `pubsub`. This closes the connections of **all the clients** in the specified class. Note that clients blocked into the `MONITOR` command are considered to belong to the `normal` class.
 * `CLIENT KILL USER username`. Closes all the connections that are authenticated with the specified [ACL](../topics/acl.md) username, however it returns an error if the username does not map to an existing ACL user.
 * `CLIENT KILL SKIPME yes/no`. By default this option is set to `yes`, that is, the client calling the command will not get killed, however setting this option to `no` will have the effect of also killing the client calling the command.
--- a/commands/client-kill.md
+++ b/commands/client-kill.md
@@ -13,7 +13,7 @@ instead of killing just by address. The following filters are available:
 
 * `CLIENT KILL ADDR ip:port`. This is exactly the same as the old three-arguments behavior.
 * `CLIENT KILL LADDR ip:port`. Kill all clients connected to specified local (bind) address.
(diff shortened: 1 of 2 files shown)


hpatro (approved):

commands/client-list.md
```diff
@@ -1,9 +1,32 @@
...
+
+- **`MAXAGE milliseconds`**: Filters the list to include only clients whose connection age (time since the client was created) is greater than or equal to the specified number of milliseconds.
+
+#### Combining Filters
+
+Filters can be combined to perform more precise searches. The command will handle multiple filters via logical AND.
+
+```bash
+CLIENT LIST TYPE normal USER admin MAXAGE 5000 ID 1234 5678
+```
```
zuiderkwast, line 29: The " Here is the meaning of the fields" below doesn't have any heading, so if we add a heading here, it makes the list of fields appear under Combining Filters. So, if we start using headings, we should have a heading for explanation of the list below and one for the filters above too.

But I don't think we need this heading. We can just keep the one sentence explaining that filters can be combined. The example can be moved to an `## Examples` section in the end of the file. This is a traditional order of sections within a man page that I think we should follow. It makes it look more like a classic manual. (See "Sections within a manual page" in the [`man-pages`](https://www.man7.org/linux/man-pages/man7/man-pages.7.html) man page.)

```suggestion
Filters can be combined to perform more precise searches. The command will handle multiple filters via logical AND.
```

... and move the example to the end of the page

````
## Examples

```bash
CLIENT LIST TYPE normal USER admin MAXAGE 5000 ID 1234 5678
```
````

**Review by hwware — APPROVED** [review]

File: commands/client-list.md [hunk 2]
```diff
@@ -1,9 +1,32 @@
...
 
-The `ID` filter only returns entries for clients with IDs matching the `client-id` arguments.
+- **`TYPE type`**: Filters the list by clients' type, where *type* is one of `normal`, `master`, `replica`, and `pubsub`.
+  > Note: Clients blocked by the `MONITOR` command belong to the `normal` class.
+
+- **`ID client-id [client-id ...]`**: Returns entries for clients with IDs matching one or more `client-id` arguments.
+
+- **`USER user`**: Filters the list to include only clients authenticated as the specified user.
+
+- **`ADDR ip:port`**: Filters the list to include only clients connected from the specified address.
```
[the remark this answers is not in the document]
zuiderkwast: I'm OK with overloaded `ADDR ip:port` and `ADDR ip`, by I'm sure if others will like it. We should have decision by the majority of the TSC about this change. Let's suggest it in the issue. valkey-io/valkey#668

What about IPv6? It already contains `:` so will it be ambiguous if `port` is optional? Or do we already require `ADDR [::1]:23456` with IPv6 with port? I don't know... We need to check.

Another suggestion for Zhao is a new filter `IP ip` or `CLIENT-IP ip`.

commands/client-kill.md
```diff
@@ -13,7 +13,7 @@ instead of killing just by address. The following filters are available:
 
 * `CLIENT KILL ADDR ip:port`. This is exactly the same as the old three-arguments behavior.
 * `CLIENT KILL LADDR ip:port`. Kill all clients connected to specified local (bind) address.
-* `CLIENT KILL ID client-id`. Allows to kill a client by its unique `ID` field. Client `ID`'s are retrieved using the `CLIENT LIST` command.
+* `CLIENT KILL ID client-id [client-id ...]`. Allows to kill a client by its unique `ID` field. Client `ID`'s can be retrieved using the `CLIENT LIST` command. The filter supports one or more `client-id` arguments
```
zuiderkwast, line 16:
```suggestion
* `CLIENT KILL ID client-id [client-id ...]`. Allows to kill a client by its unique `ID` field. Client `ID`'s can be retrieved using the `CLIENT LIST` command. The filter supports one or more `client-id` arguments.
```

zuiderkwast (approved):
LGTM.

Just one missing dot in the end of a sentence. :)

commands/client-list.md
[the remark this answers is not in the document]
sarthakaggarwal97: > Another suggestion by Zhao is a new filter IP ip or CLIENT-IP ip.

yeah, even I was thinking this. @hwware also gave some suggestions on https://github.com/valkey-io/valkey/issues/668, let me respond back with the support for current PR and things we can add.

commands/client-kill.md
sarthakaggarwal97: made the change :)

PR #226, opened 2025-02-06
Improve docs for rolling upgrade of a cluster

zuiderkwast:
Upgrading a cluster is described very briefly in the Cluster tutorial. This PR expands this with a more thorough step-by-step procedure for a rolling upgrade.

If there are better ways e.g. to check if a replica is ready, please let me know.

### DIFF
--- a/topics/cluster-tutorial.md
+++ b/topics/cluster-tutorial.md
@@ -726,12 +726,12 @@ This is what you see in the replica log when you perform a manual failover:
     # Starting a failover election for epoch 7545.
     # Failover election won: I'm the new primary.
 
-Basically clients connected to the primary we are failing over are stopped.
-At the same time the primary sends its replication offset to the replica, that
+Clients sending commands to the primary are blocked during the failover.
+When the primary sends its replication offset to the replica, the replica
 waits to reach the offset on its side. When the replication offset is reached,
 the failover starts, and the old primary is informed about the configuration
-switch. When the clients are unblocked on the old primary, they are redirected
-to the new primary.
+switch. When the switch is complete, the clients are unblocked on the old
+primary and they are redirected to the new primary.
 
 **Note:**
 To promote a replica to primary, it must first be known as a replica by a majority of the primaries in the cluster.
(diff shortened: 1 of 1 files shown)


nastena1606 (commented):
LGTM, just a side node. Perhaps it makes sense to clarify what nodes to start the rolling upgrade from. What do you think?

zuiderkwast:
> what nodes to start the rolling upgrade from

What do you mean? Which nodes to upgrade first?

nastena1606:
> > what nodes to start the rolling upgrade from
> 
> What do you mean? Which nodes to upgrade first?

Yes

zuiderkwast:
> > > what nodes to start the rolling upgrade from
> > 
> > 
> > What do you mean? Which nodes to upgrade first?
> 
> Yes

I did, didn't I? Add new replicas (optional) and upgrade the old replicas are the first two steps. Then upgrade the primary, the following steps. What am I missing?

madolson (commented):
Minor comments, but overall looks correct.

topics/cluster-tutorial.md
```diff
@@ -928,16 +928,53 @@ it with an updated version of Valkey. If there are clients scaling reads using
...
+   described under [Add a new node as a replica](#add-a-new-node-as-a-replica).
+
+2. Upgrade the existing replicas you want to keep by restarting them with the
+   updated version of Valkey. If you're replacing all the old nodes with new
+   nodes, you can skip this step.
+
+3. Select one of the upgraded replicas to be the new primary. Wait until this
+   replica has caught up the replication offset with the primary. You can use
+   [`INFO REPLICATION`](../commands/info.md) and check for the line
+   `master_link_status:up` to be present. This indicates that the sync with the
```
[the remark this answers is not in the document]
zuiderkwast: Thanks! (I took `master_link_status:up` from our test suite's `proc wait_for_sync`. :smile:)

How do you suggest we describe this? Like this?

```TCL
proc wait_for_ofs_sync {r1 r2} {
    wait_for_condition 50 100 {
        [status $r1 master_repl_offset] eq [status $r2 master_repl_offset]
    } else {
        fail "replica offset didn't match in time"
    }
}
```

Or can you just just compare `master_repl_offset` to `slave_repl_offset` reported by the replica?

If there is a constant stream of writes, the offsets may never be completely equal, right? The admin should just try to minimize the diff but not require them to be exactly equal?

topics/cluster-tutorial.md
```diff
@@ -928,16 +928,53 @@ it with an updated version of Valkey. If there are clients scaling reads using
...
+   primary is complete.
+
+4. Check that the new replica is known by all nodes in the cluster, or at least
+   by the primaries in the cluster. You can send [`CLUSTER
+   NODES`](../commands/cluster-nodes.md) to each of the nodes in the cluster and
+   check that they all are aware of the new node. Wait for some time and repeat
+   the check if necessary.
```
hpatro, line 959: Might be handy to to build it in as a utility in the `valkey-cli` to perform this check.

topics/cluster-tutorial.md
```diff
@@ -928,16 +928,53 @@ it with an updated version of Valkey. If there are clients scaling reads using
...
+A complete rolling upgrade of all nodes in a cluster can be performed by
+repeating the following procedure for each shard (a primary and its replicas):
+
+1. Add one or more upgraded nodes as new replicas to the primary. This step is
+   optional but it ensures that the number of replicas is not compromised during
+   the rolling upgrade. An alternative is to upgrade one replica at a time and
+   have fewer replicas online during the upgrade. To add a new node, use
+   [`CLUSTER MEET`](../commands/cluster-meet.md) and [`CLUSTER
+   REPLICATE`](../commands/cluster-replicate.md) or use `valkey-cli` as
+   described under [Add a new node as a replica](#add-a-new-node-as-a-replica).
```
hpatro, line 943: Should we structure this as approach 1 (Existing set of nodes)/approach 2 (Additional node for better availability guarantees)? Or let's add a newline for the alternative suggestion.

topics/cluster-tutorial.md
```diff
@@ -928,16 +928,53 @@ it with an updated version of Valkey. If there are clients scaling reads using
...
+5. Trigger a manual failover by sending [`CLUSTER
+   FAILOVER`](../commands/cluster-failover.md) to the replica node selected to
+   become the new primary. See the [Manual failover](#manual-failover) section
+   in this document for more information.
+
+6. Wait for the failover to complete. To check, you can use
+   [`ROLE`](../commands/role.md), [`INFO REPLICATION`](../commands/info.md)
+   (which indicates “role:master” after successful failover) or [`CLUSTER
+   NODES`](../commands/cluster-nodes.md) to verify that the state of the cluster
+   has changed sometime after the command was sent.
```
hpatro, line 970: nit: `sometime after` feels a bit weird. `shortly after` ?

topics/cluster-tutorial.md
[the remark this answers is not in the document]
zuiderkwast: So, compare `master_repl_offset` by querying both nodes? The `slave_repl_offset` isn't useful, is it? We don't seem to use it in our test suites much.

topics/cluster-tutorial.md
zuiderkwast: Yes, that'd be good, as a reference implementation. The first step is docs though IMHO.

Even better would be to have an official open source kubernetes operator to orchestrate things like rolling upgrade and scaling, even autoscaling, by deploying new pods as necessary.

topics/cluster-tutorial.md
zuiderkwast: :+1: Thanks, I'm not a native English speaker.

topics/cluster-tutorial.md
zuiderkwast: Do you mean repeat the full 7 steps for the 2 approaches?

I'd prefer to keep them unified so I'll just add a newline. You could actually do any mix of the to approaches (add some, upgrade some, remove some). We replace all the nodes and remove all the old ones. I think it depends whether you use VMs or pods, and whether you use persistence or not.

zuiderkwast: I've added a blank line between the alternatives here.

topics/cluster-tutorial.md
hpatro: Definitely.

topics/cluster-tutorial.md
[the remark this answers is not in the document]
zuiderkwast: I added this:

> After the initial full sync, the replica might still lag behind in
   replication. Send `INFO REPLICATION` to the primary and the replica and
   compare the field `master_repl_offset` returned by both nodes. If the offsets
   match, it means that all writes have been replicated. However, if the primary
   receives a constant stream of writes, it's possible that the offsets will
   never be equal. In this step, you can accept a small difference. It's usually
   enough to wait for some seconds to minimize the difference.

Good enough?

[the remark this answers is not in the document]
enjoy-binbin: > At one point we had a "replication_lag" metric PR, that showed the time delta of how far the replica was behind which was a useful metric to track, I think that might have gotten silently dropped during the fork.

aha, i had some memory on this one, you means this one? offset_lag  https://github.com/redis/redis/pull/13036?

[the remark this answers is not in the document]
madolson: That one, and this one was also floating around https://github.com/redis/redis/pull/13106.

madolson (approved):

topics/cluster-tutorial.md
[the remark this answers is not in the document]
zuiderkwast: I wrote here that you should check `master_repl_offset` by sending INFO to both nodes, but you can also just use the INFO of the primary and compare the lines like these ones?

    slave0:ip=127.0.0.1,port=30004,state=wait_bgsave,offset=0

This is the ACKed offset, right? It seems easier for an admin (control plane) to just query the primary repeatedly than to query both nodes. WDYT?

PR #230, opened 2025-02-11
Update the active expire algorithm on the EXPIRE command page

### LINKED ISSUE #185: EXPIRE command docs incorrect says active expire uses random sampling
The EXPIRE command manual (`commands/expire.md`) says that active expiry uses random sampling. This was changed some years ago (Redis 6?) to a scan based approach, i.e. an incremental sweep over all keys with expire.

This is the incorrect part of the page:

```
   How Valkey expires keys
       Valkey  keys  are  expired  in two ways: a passive way, and an active
       way.

       A key is passively expired simply when some client  tries  to  access
       it, and the key is found to be timed out.

       Of course this is not enough as there are expired keys that will nev‐
       er  be accessed again.  These keys should be expired anyway, so peri‐
       odically Valkey tests a few keys at random among keys with an  expire
       set.  All the keys that are already expired are deleted from the key‐
       space.

       Specifically this is what Valkey does 10 times per second:

       1. Test  20  random  keys from the set of keys with an associated ex‐
          pire.

       2. Delete all the keys found expired.

       3. If more than 25% of keys were expired, start again from step 1.

       This is a trivial probabilistic algorithm, basically  the  assumption
       is  that  our sample is representative of the whole key space, and we
       continue to expire until the percentage of keys that are likely to be
       expired is under 25%

       This means that at any given moment the maximum amount  of  keys  al‐
       ready  expired that are using memory is at max equal to max amount of
       write operations per second divided by 4.
```

nastena1606:
Change the description of the active expire algorithm to describe the current scan-based approach instead of the old no longer used random sampling approach.

Fixes #185

### DIFF
--- a/commands/expire.md
+++ b/commands/expire.md
@@ -159,10 +159,10 @@ Changing the `active-expire-effort` value results in a lower percentage of expir
 
 To calculate the new values, use the following formulas:
 
-* `ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP + (ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP / 4 * effort)`
-* `ACTIVE_EXPIRE_CYCLE_FAST_DURATION + (ACTIVE_EXPIRE_CYCLE_FAST_DURATION / 4 * effort)`
-* `ACTIVE_EXPIRE_CYCLE_SLOW_TIME_PERC + (2 * effort)`
-* `ACTIVE_EXPIRE_CYCLE_ACCEPTABLE_STALE - effort`
+* `ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP + (ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP / 4 * (effort - 1))`
+* `ACTIVE_EXPIRE_CYCLE_FAST_DURATION + (ACTIVE_EXPIRE_CYCLE_FAST_DURATION / 4 * (effort - 1))`
+* `ACTIVE_EXPIRE_CYCLE_SLOW_TIME_PERC + (2 * (effort - 1))`
+* `ACTIVE_EXPIRE_CYCLE_ACCEPTABLE_STALE - (effort - 1)`
 
 where `ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP`, `ACTIVE_EXPIRE_CYCLE_FAST_DURATION`, `ACTIVE_EXPIRE_CYCLE_SLOW_TIME_PERC`, and `ACTIVE_EXPIRE_CYCLE_ACCEPTABLE_STALE` are the base values, and `effort` is calculated as the specified `active-expire-effort` value minus 1.
 
--- a/commands/expire.md
+++ b/commands/expire.md
@@ -138,32 +138,34 @@ you set a key with a time to live of 1000 seconds, and then set your computer
 time 2000 seconds in the future, the key will be expired immediately, instead of
 lasting for 1000 seconds.
 
-## How Valkey expires keys
+## How Valkey reclaims expired keys
 
-Valkey keys are expired in two ways: a passive way, and an active way.
+Valkey reclaims expired keys in two ways: on access and in the background in what is called the "active expire key" cycles. On access expiration is when a client tries to access a key with the expiration time which is found to be timed out. Such a key is deleted on this access attempt.
 
-A key is passively expired simply when some client tries to access it, and the
-key is found to be timed out.
+Relying solely on the on access expiration only is not enough because there are expired keys that will never be accessed again. To address this, Valkey uses the background expiration known as the "active expire key" effort. Valkey slowly and interactively scans the keyspace to identify and reclaim expired keys. This slow cycle is the main way to collect expired keys and operates with the server's hertz frequency (usually 10 hertz).
 
-Of course this is not enough as there are expired keys that will never be
-accessed again.
-These keys should be expired anyway, so periodically Valkey tests a few keys at
-random among keys with an expire set.
-All the keys that are already expired are deleted from the keyspace.
+During the "slow cycle", Valkey scans 20 keys per database loop. It tolerates having not more than 10% of the expired keys in the memory and tries to use a maximum of 25% CPU power. These default values are adjusted if the user changes the active expire key effort configuration. 
 
-Specifically this is what Valkey does 10 times per second:
+If the number of expired keys remains high after the slow cycle, the active expire key effort transitions into the "fast cycle", trying to do less work but more often. The fast cycle runs no longer than 1000 microseconds and repeats at the same interval. During the fast cycle, the check of every database is interrupted once the number of already expired keys in the database is estimated to be lower than 10%. This is done to avoid doing too much work to gain too little memory. 
 
-1. Test 20 random keys from the set of keys with an associated expire.
-2. Delete all the keys found expired.
-3. If more than 25% of keys were expired, start again from step 1.
+You can modify the active expire key effort with the `active-expire-effort` parameter in the configuration file up to the maximum value of `10`. The default `active-expire-effort` value is `1`, and it is described by the following base values:
 
-This is a trivial probabilistic algorithm, basically the assumption is that our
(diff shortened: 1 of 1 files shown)


commands/expire.md
```diff
@@ -138,32 +138,34 @@ you set a key with a time to live of 1000 seconds, and then set your computer
...
+* `ACTIVE_EXPIRE_CYCLE_ACCEPTABLE_STALE` = 10 – The maximum % of expired keys to tolerate in memory. 
+
+Changing the `active-expire-effort` value results in a lower percentage of expired keys tolerated in memory. However, it will lead to longer cycles and increased CPU usage, which may introduce latency.
+
+To calculate the new values, use the following formulas:
+
+* `ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP + (ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP / 4 * effort)`
+* `ACTIVE_EXPIRE_CYCLE_FAST_DURATION + (ACTIVE_EXPIRE_CYCLE_FAST_DURATION / 4 * effort)`
+* `ACTIVE_EXPIRE_CYCLE_SLOW_TIME_PERC + (2 * effort)`
+* `ACTIVE_EXPIRE_CYCLE_ACCEPTABLE_STALE - effort`
```
zuiderkwast, line 165: Let's just put in `(expire - 1)` here?

```suggestion
* `ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP + (ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP / 4 * (effort - 1))`
* `ACTIVE_EXPIRE_CYCLE_FAST_DURATION + (ACTIVE_EXPIRE_CYCLE_FAST_DURATION / 4 * (effort - 1))`
* `ACTIVE_EXPIRE_CYCLE_SLOW_TIME_PERC + (2 * (effort - 1))`
* `ACTIVE_EXPIRE_CYCLE_ACCEPTABLE_STALE - (effort - 1)`
```

I think you missed this line in the activeExpireCycle function:

```c
    unsigned long effort = server.active_expire_effort - 1, /* Rescale from 0 to 9. */
```

commands/expire.md
```diff
@@ -138,32 +138,34 @@ you set a key with a time to live of 1000 seconds, and then set your computer
...
+
+where `ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP`, `ACTIVE_EXPIRE_CYCLE_FAST_DURATION`, `ACTIVE_EXPIRE_CYCLE_SLOW_TIME_PERC`, and `ACTIVE_EXPIRE_CYCLE_ACCEPTABLE_STALE` are the base values, and `effort` is calculated as the specified `active-expire-effort` value minus 1.
```
zuiderkwast, line 167: If we use `(expire - 1)` in the forumlas above, we can just say that `expire` is the configured `active-expire-effort`. It makes it easier to follow I think.

zuiderkwast (commented):
Great work! Thanks!

commands/expire.md
```diff
@@ -138,32 +138,34 @@ you set a key with a time to live of 1000 seconds, and then set your computer
...
-A key is passively expired simply when some client tries to access it, and the
-key is found to be timed out.
+Relying solely on the on access expiration only is not enough because there are expired keys that will never be accessed again. To address this, Valkey uses the background expiration known as the "active expire key" effort. Valkey slowly and interactively scans the keyspace to identify and reclaim expired keys. This slow cycle is the main way to collect expired keys and operates with the server's hertz frequency (usually 10 hertz).
 
-Of course this is not enough as there are expired keys that will never be
-accessed again.
-These keys should be expired anyway, so periodically Valkey tests a few keys at
-random among keys with an expire set.
-All the keys that are already expired are deleted from the keyspace.
+During the "slow cycle", Valkey scans 20 keys per database loop. It tolerates having not more than 10% of the expired keys in the memory and tries to use a maximum of 25% CPU power. These default values are adjusted if the user changes the active expire key effort configuration.
```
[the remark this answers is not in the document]
zuiderkwast: I mean we can skip everything related to `ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP`.

`CRON_DBS_PER_CALL` is 16 (defined in server.h) and it's the number of databases that are scanned in each cron cycle. But many users use only one database so in many cases it's irrelevant. In cluster mode, it's impossible to use more than one database. That's why I think this is mostly too internal. It's basically a safeguard to make sure the server doesn't get stuck if there are too many databases, which is an edge case.

commands/expire.md
```diff
@@ -138,32 +138,34 @@ you set a key with a time to live of 1000 seconds, and then set your computer
...
 
-## How Valkey expires keys
+## How Valkey reclaims expired keys
 
-Valkey keys are expired in two ways: a passive way, and an active way.
+Valkey reclaims expired keys in two ways: on access and in the background in what is called the "active expire key" cycles. On access expiration is when a client tries to access a key with the expiration time which is found to be timed out. Such a key is deleted on this access attempt.
 
-A key is passively expired simply when some client tries to access it, and the
-key is found to be timed out.
+Relying solely on the on access expiration only is not enough because there are expired keys that will never be accessed again. To address this, Valkey uses the background expiration known as the "active expire key" effort. Valkey slowly and interactively scans the keyspace to identify and reclaim expired keys. This slow cycle is the main way to collect expired keys and operates with the server's hertz frequency (usually 10 hertz).
```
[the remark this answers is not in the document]
nastena1606: Well, then we should also correct the conf file description: https://github.com/valkey-io/valkey/blob/8.0.2/valkey.conf#L1234
Will remove "interactively" completely to avoid confusion

commands/expire.md
nastena1606: Agree. Updated

commands/expire.md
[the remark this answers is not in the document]
nastena1606: > You've seen this code comment in expire.c right?
> 
> https://github.com/valkey-io/valkey/blob/8.0.2/src/expire.c#L74
> 
I did:) 
> If not, I have no idea how you would know all this. 😄
> 
> Here, we use the term _database loop_ without first explaining what it means. I think it's too technical. I would just skip everything about _database loop_. If anyone wants to understand this level of detail, they should look up the source code IMO.

Agree, I removed it from the description.
> 
> Also the slow and the fast cycle are quite some technical details but these are explained below. But I think we should first explain something more high level about the algorithm, like this from the code comment:
Updated.

commands/expire.md
```diff
@@ -138,32 +138,36 @@ you set a key with a time to live of 1000 seconds, and then set your computer
...
-2. Delete all the keys found expired.
-3. If more than 25% of keys were expired, start again from step 1.
+If the number of expired keys remains high after the slow cycle, the active expire key effort transitions into the "fast cycle", trying to do less work but more often. The fast cycle runs no longer than 1000 microseconds and repeats at the same interval. During the fast cycle, the check of every database is interrupted once the number of already expired keys in the database is estimated to be lower than 10%. This is done to avoid doing too much work to gain too little memory. 
 
-This is a trivial probabilistic algorithm, basically the assumption is that our
-sample is representative of the whole key space, and we continue to expire until
-the percentage of keys that are likely to be expired is under 25%
+You can modify the active expire key effort with the `active-expire-effort` parameter in the configuration file up to the maximum value of `10`. The default `active-expire-effort` value is `1`, and it is described by the following base values:
+
+* `ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP` = 20 - The number of keys for each DB loop.
```
zuiderkwast, line 155: I was suggesting removing this variable too (because it's about the database loop that we didn't explain), but we can keep it. Even if it's not really explained what it doesn, it can give some clue to some readers, maybe.

nastena1606: My concern is that if we remove the variable, then all formulas must go away, too. Then there should be another way to explain what happens after modifying the `active-expire-effort`

zuiderkwast: No, ACTIVE_EXPIRE_CYCLE_KEYS_PER_LOOP is only part of the first formula. We can keep the other three.

zuiderkwast: But let's keep it. :)

PR #231, opened 2025-02-14
Update lists.md

yoshuacas:
The example showing that RPUSH and LPUSH are variadic commands is incorrectly returning data from the wrong list name. The list name should be bikes:repairs, but the current example uses LRANGE on mylist.


madolson (approved):
Thanks! Sorry for missing this earlier, appreciate the contribution.

PR #238, opened 2025-02-26
Add cluster-manual-failover-timeout desc in CLUSTER FAILOVER command page

enjoy-binbin:
See https://github.com/valkey-io/valkey/pull/1690

### DIFF
--- a/commands/cluster-failover.md
+++ b/commands/cluster-failover.md
@@ -16,6 +16,11 @@ This way clients are moved away from the old primary to the new primary
 atomically and only when the replica that is turning into the new primary
 has processed all of the replication stream from the old primary.
 
+The default failover timeout is 5000ms, it is possible to configure the timeout
+via `cluster-manual-failover-timeout` configuration item (added in Valkey 8.1)
+and decide how long the primary will pause in the worst case scenario,  i.e. the
+manual failover timed out due to the insufficient votes.
+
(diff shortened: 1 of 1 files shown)


commands/cluster-failover.md
```diff
@@ -16,6 +16,11 @@ This way clients are moved away from the old primary to the new primary
 atomically and only when the replica that is turning into the new primary
 has processed all of the replication stream from the old primary.
 
+The default failover timeout is 5000ms, it is possible to configure the timeout
+via `cluster-manual-failover-timeout` configuration item (added in Valkey 8.1)
+and decide how long the primary will pause in the worst case scenario,  i.e. the
+manual failover timed out due to the insufficient votes.
```
madolson, line 22:
```suggestion
There is a built-in timeout for the primary to resume processing queries if the
replica does not catch quickly enough or is unable to secure the votes to take over.
The default failover timeout is 5000ms, it is possible to configure the timeout
via `cluster-manual-failover-timeout` configuration (added in Valkey 8.1)
and decide how long the primary will pause in the worst case scenario, i.e. the
manual failover timed out due to the insufficient votes.
```
Looks good, but I don't see context on the timeout, so maybe we can add one sentence for that.

enjoy-binbin: good sentence!

PR #239, opened 2025-02-27
Add workflow for deploying website after updating the documentation

madolson:
This will allow the website to automatically get updated when we update the doc pages is updated. I setup a repository action on the website github, which allows incoming events to trigger the build and deploy action. 

Some notes about security. The credential being used has write permission the website repository, so to compensate I locked down all the branches to require PRs, which should prevent any malicious attackers if they get access to the credentials. The downside is no more directly pushing to the website, which I think is an OK tradeoff. 

This PR is dependent on this PR getting merged first: https://github.com/valkey-io/valkey-io.github.io/pull/214.

### DIFF
--- a/.github/workflows/trigger-build.yml
+++ b/.github/workflows/trigger-build.yml
@@ -1,18 +1,16 @@
-name: Trigger master website deploy
+name: Trigger website deploy
 on:
   push:
     branches:
-      - master
+      - main
 
 jobs:
-  trigger:
+  deploy:
     runs-on: ubuntu-latest
     steps:
-      - run: |
-          echo "'$DATA'" | xargs \
-          curl \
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-build.yml
```diff
@@ -1,18 +1,16 @@
-name: Trigger master website deploy
+name: Trigger website deploy
 on:
   push:
     branches:
```
madolson, line 4: We could scope down the file paths that trigger this, but I decided to leave this as is for now. The rate of change on the documentation repository isn't that high.

zuiderkwast (approved):
Thank you!

Very nice. It's very satisfying to see changes deployed immediately when merging a PR. It's nice to see for the contributor too.

Requiring PRs on the website repo is a good idea regardless.

PR #240, opened 2025-02-27
Update Cluster Bus Port out of range error message for Cluster Meet command

hwware:
This PR is related to PR https://github.com/valkey-io/valkey/pull/1686

### DIFF
--- a/resp2_replies.json
+++ b/resp2_replies.json
@@ -259,7 +259,7 @@
     "[Array reply](../topics/protocol.md#arrays): an array of maps where each map contains various attributes and their values of a cluster link."
   ],
   "CLUSTER MEET": [
-    "[Simple string reply](../topics/protocol.md#simple-strings): `OK` if the command was successful. If the address or port specified are invalid an error is returned."
+    "[Simple string reply](../topics/protocol.md#simple-strings): `OK` if the command was successful. If the port or cluster bus port number is out of range, an error is returned. If the ip address is invalid, Invalid node address message is returned."
   ],
   "CLUSTER MYID": [
     "[Bulk string reply](../topics/protocol.md#bulk-strings): the node ID."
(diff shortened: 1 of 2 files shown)


resp2_replies.json
```diff
@@ -259,7 +259,7 @@
     "[Array reply](../topics/protocol.md#arrays): an array of maps where each map contains various attributes and their values of a cluster link."
   ],
   "CLUSTER MEET": [
-    "[Simple string reply](../topics/protocol.md#simple-strings): `OK` if the command was successful. If the address or port specified are invalid an error is returned."
+    "[Simple string reply](../topics/protocol.md#simple-strings): `OK` if the command was successful. If the port or cluster bus port number is out of range, an error is returned. If the ip address is invalid, Invalid node address message is returned."
```
zuiderkwast, line 262: "Invalid node address message" is an error too.

Do we need to specify document the exact error message here? I think we usually don't document it.

What do you think about this?

```suggestion
    "[Simple string reply](../topics/protocol.md#simple-strings): `OK` if the command was successful. If the port or cluster bus port number is out of range, or if an invalid address is specified, an error is returned."
```

hwware: Address your comments in 2 files. Please take a look, Thanks

resp3_replies.json
```diff
@@ -259,7 +259,7 @@
     "[Array reply](../topics/protocol.md#arrays): an array of [Map reply](../topics/protocol.md#maps) where each map contains various attributes and their values of a cluster link."
   ],
   "CLUSTER MEET": [
-    "[Simple string reply](../topics/protocol.md#simple-strings): `OK` if the command was successful. If the address or port specified are invalid an error is returned."
+    "[Simple string reply](../topics/protocol.md#simple-strings): `OK` if the command was successful.If the port or cluster bus port number is out of range, or if an invalid address is specified, an error is returned."
```
zuiderkwast, line 262: Looks good, but a space was removed by mistake.

```suggestion
    "[Simple string reply](../topics/protocol.md#simple-strings): `OK` if the command was successful. If the port or cluster bus port number is out of range, or if an invalid address is specified, an error is returned."
```

zuiderkwast (approved):

PR #245, opened 2025-03-10
fix glide-node installation, update iovalkey version

liorsve:
This PR - 
- Fixes the glide-node npm installation instructions
- Updates iovalkey to the latest version


clients/node.js/iovalkey.json
```diff
@@ -3,8 +3,8 @@
     "description": "A robust, performance-focused and full-featured Redis client for Node.js.",
     "repo":"https://github.com/valkey-io/iovalkey",
     "installation": "npm install iovalkey",
-    "version":"v0.3.0",
-    "version_released":"2025-01-23",
+    "version":"v0.3.1",
+    "version_released":"2025-03-10",
```
madolson, line 7: @liorsve Can we investigate automating this? When we are deploying the website, we should see if we can get the latest version from github for these values.

madolson (approved):

clients/node.js/iovalkey.json
liorsve: @madolson I think it should be easy enough using github API. But - I believe the whole reason for including the version and date in the first place (mainly suggested by @zuiderkwast) was to indicate when the data on the website was last updated, so even if it would get outdated it won't mislead. So having it automated may get us to a situation where the data is not correct for the version specified. 
LMK what you think, if automating is still preferred I can get to working on it.

zuiderkwast: Yeah, when we update the version, we should also update the supported features and other info to be accurate for the version described.

madolson: Hmm, are we not concerned people will think to use the wrong version because it wasn't updated? It seems like extra work to make client developers update this page when nothing else needs to get updated.

zuiderkwast: Yes, that's pros and cons. I think the clients' own release notes, readmes and similar need to be the primary source of information. We just provide an overview and comparison. Users need to be aware of this.

Maybe we should add a disclaimer to this page?

madolson: We could source the information from a JSON file in their own repository if they want to provide it. Then they don't need to come update this repository with the information.

liorsve: Sorry for the delay 
@madolson @zuiderkwast Are you suggesting that we source only the version information from each client’s own repo, or all of the JSON data?

On one hand, putting all the version info in each individual repo could make the site more fragile, since any small change in format might break the page and there won't be any website maintainers supervision over this. On the other hand, if the sole purpose of this version data is to be used by the clients page on valkey.io, it seems a bit odd to source some of that info from the repos themselves and some from valkey-doc. It would also require coordination across all maintainers, which might take time, and I’m not sure it would actually improve the accuracy of the version data on the site.

Maybe this is something we should raise in the next clients maintainers meeting - If there’s general agreement to automate this via the github API and include a disclaimer encouraging users to consult the repos directly for the most up-to-date info, maybe that might still be the better approach?

PR #250, opened 2025-03-21
Fix stream names in the streams intro

gkuznets:
Fixing inconsistent example stream names in the streams intro.


madolson (approved):
Makes sense

PR #264, opened 2025-04-16
Amend note about valkey-cli about testing RDMA

zuiderkwast:
Incorrect information about valkey-cli using IPoIB was introduced in #217.


zuiderkwast:
@pizhenwei @z-koki

PR #267, opened 2025-04-24
Correct RPUSHX command doc

hwware:
Related to PR https://github.com/valkey-io/valkey/pull/2002


zuiderkwast (approved):
OK, but this JSON file is not used in this repo. We should delete it.

hpatro:
> OK, but this JSON file is not used in this repo. We should delete it.

was trying to understand why do we maintain this file. If it's unused, I would also recommend cleaning it up rather than merging this PR.

zuiderkwast:
It was used in redis website. Valkey never used it. Nobody has deleted it so far. Feel free.

PR #269, opened 2025-05-01
Adds a workflow to test the man page build and install.

roshkhatri:
We do not have a wf that tests the man page build and install for the docs.

This adds a workflow to test the man page build and install.

### DIFF
--- a/.github/workflows/man-page-build-install.yml
+++ b/.github/workflows/man-page-build-install.yml
@@ -2,12 +2,10 @@ name: Man Page Build and Install
 
 on:
   push:
-    # branches:
-    #   - main
   pull_request:
 
 jobs:
-  build-and-install:
+  man-pages-build-install:
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/man-page-build-install.yml
+++ b/.github/workflows/man-page-build-install.yml
@@ -5,7 +5,59 @@ on:
   pull_request:
 
 jobs:
-  man-pages-build-install:
+  man-pages-build-install-without-modules:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout valkey-doc
+      uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+      with:
+        path: valkey-doc
+
+    - name: Checkout valkey
+      uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+      with:
+        repository: valkey-io/valkey
+        path: valkey
+
+    - name: Set up Python
+      uses: actions/setup-python@7f4fc3e22c37d6ff65e88745f38bd3157c663f7c # v4.9.1
+      with:
+        python-version: '3.x'
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y make pandoc python3-pip
+        pip3 install pyyaml
+
+    - name: Build and install man pages
+      working-directory: valkey-doc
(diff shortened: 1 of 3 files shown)


.github/workflows/man-page-build-install.yml
```diff
@@ -0,0 +1,72 @@
...
+        if [ -z "$(ls -A _build/man)" ]; then
+          echo "Error: No man pages were generated in '_build/man'!"
+          exit 1
+        fi
+        echo "Man pages successfully built in _build/man/"
+
+    - name: Install man pages
+      working-directory: valkey-doc
+      run: |
+        sudo make install INSTALL_MAN_DIR=/usr/local/share/man
```
madolson, line 72: @zuiderkwast Is there anything else you would like to validate to make sure we don't silently break the man page build?

.github/workflows/man-page-build-install.yml
```diff
@@ -0,0 +1,72 @@
...
+        sudo apt-get install -y make pandoc python3-pip
+        pip3 install pyyaml
+
+    - name: Build and install man pages
+      working-directory: valkey-doc
+      run: |
+        make VALKEY_ROOT=$GITHUB_WORKSPACE/valkey \
+             VALKEY_BLOOM_ROOT=$GITHUB_WORKSPACE/valkey-bloom \
+             VALKEY_JSON_ROOT=$GITHUB_WORKSPACE/valkey-json \
+              -j;
```
zuiderkwast, line 52: It'd be nice to check that it can build without the Bloom and JSON modules too.

Maybe add a build step without those first, and then build with them?

.github/workflows/man-page-build-install.yml
```diff
@@ -0,0 +1,72 @@
+name: Man Page Build and Install
+
+on:
+  push:
+  pull_request:
+
```
zuiderkwast, line 6: We can prevent writes to the repos. An extra security thing.

```suggestion

permissions:
  contents: read

```

zuiderkwast (commented):
Awesome, thank you Roshan!

.github/workflows/man-page-build-install.yml
zuiderkwast: I don't know. Checking that it can be built and installed is a good start.

I just wish it would be used more by distros and included when you install valkey. AFAIK only Fedora packages valkey-doc and users need to install it separately.

madolson: Adoption always takes time ¯\_(ツ)_/¯

.github/workflows/man-page-build-install.yml
madolson: Is this worth adding everywhere? Outside of a couple of our automation workflows, everything does reads.

zuiderkwast: Yeah, I guess it's from some list of best-practices. Maybe OpenSSF? I can't see it now but this list has a lot of ideas: https://best.openssf.org/SCM-BestPractices/github/

.github/workflows/man-page-build-install.yml
zuiderkwast: Hehe, it caught something. :)

```
Makefile:40: *** Please provide the VALKEY_JSON_ROOT variable pointing to the valkey-json source code.  Stop.
Valkey bloom variable pointed to nothing, skipping bloom filter commands
```

roshkhatri: Yes, I fixed this in the update haha

zuiderkwast (approved):

PR #286, opened 2025-05-29
Update resp2_replies.json about JSON.CLEAR

momadvisor:
fix  a wrong and add a new

### DIFF
--- a/resp2_replies.json
+++ b/resp2_replies.json
@@ -748,7 +748,7 @@
     "  * Returns 0 if clearing an empty array or object.",
     "  * Returns 0 if clearing a non-container value except JSON string.",
     "  * Returns 0 if no array or object value is located by the path.",
-    "  * Returns 1 if clearing a JSON string value regardless of the length of the JSON string value.",
+    "  * Always Returns 1 if clearing a JSON string value regardless of the length of the JSON string value including empty string.",
     ""
   ],
   "JSON.DEBUG": [
--- a/resp2_replies.json
+++ b/resp2_replies.json
@@ -746,8 +746,9 @@
     "* [Integer reply](../topics/protocol.md#integers):",
     "  * The number of containers cleared.",
     "  * Returns 0 if clearing an empty array or object.",
-    "  * Returns 0 if clearing a non-container value.",
+    "  * Returns 0 if clearing a non-container value except JSON string.",
     "  * Returns 0 if no array or object value is located by the path.",
+    "  * Returns 1 if clearing a JSON string value regardless of the length of the JSON string value.",
     ""
   ],
   "JSON.DEBUG": [


resp2_replies.json
```diff
@@ -746,8 +746,9 @@
     "* [Integer reply](../topics/protocol.md#integers):",
     "  * The number of containers cleared.",
     "  * Returns 0 if clearing an empty array or object.",
-    "  * Returns 0 if clearing a non-container value.",
+    "  * Returns 0 if clearing a non-container value except JSON string.",
     "  * Returns 0 if no array or object value is located by the path.",
+    "  * Always Returns 1 if clearing a JSON string value regardless of the length of the JSON string value including empty string.",
```
sarthakaggarwal97, line 751:
```suggestion
    "  *Returns 1 when clearing a JSON value.",
```

resp2_replies.json
```diff
@@ -817,8 +746,9 @@
     "* [Integer reply](../topics/protocol.md#integers):",
     "  * The number of containers cleared.",
     "  * Returns 0 if clearing an empty array or object.",
```
roshkhatri, line 748:
```suggestion
"  * The number of values cleared.",
"  * A value is counted only when clearing changes it: a non-empty array or object is emptied, a non-empty string is set to empty, a non-zero number is set to 0, and true is set to false.",
"  * Returns 0 for a value already empty, zero, false, or null.",
```

resp2_replies.json
roshkhatri: it return the count of values, so this is incorrect.

resp2_replies.json
```diff
@@ -817,8 +746,9 @@
...
-    "  * Returns 0 if clearing a non-container value.",
+    "  * Returns 0 if clearing a non-container value except JSON string.",
     "  * Returns 0 if no array or object value is located by the path.",
+    "  * Always Returns 1 if clearing a JSON string value regardless of the length of the JSON string value including empty string.",
```
roshkhatri, line 751:
```suggestion
```

roshkhatri (commented):
We will also need to edit the resp3_replies.json

`dom_clear`:
```
} else if (v.first->IsBool()) {
    if (v.first->IsTrue()) {
        v.first->SetBool(false);          // true  --> false   (counted)
        elements_cleared++;
    }
} else if (v.first->IsInt()) {
    if (v.first->GetInt() != 0) {
        v.first->SetInt(0);               // non-zero int --> 0  (counted)
        elements_cleared++;
    }
} else if (v.first->IsInt64())  { if (... != 0) { SetInt64(0);  ... } }
} else if (v.first->IsUint())   { if (... != 0) { SetUint(0);   ... } }
} else if (v.first->IsUint64()) { if (... != 0) { SetUint64(0); ... } }
} else if (v.first->IsDouble()) {
    if (GetDouble() < 0.0 || GetDouble() > 0.0) {
        v.first->SetDouble("0.0", 3, allocator);  // non-zero double --> 0.0 (counted)
        elements_cleared++;
    }
}
```

resp2_replies.json
```diff
@@ -746,8 +746,9 @@
     "* [Integer reply](../topics/protocol.md#integers):",
     "  * The number of containers cleared.",
     "  * Returns 0 if clearing an empty array or object.",
-    "  * Returns 0 if clearing a non-container value.",
+    "  * Returns 0 if clearing a non-container value except JSON string.",
```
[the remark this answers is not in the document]
roshkhatri:
```suggestion
    "  * Returns 0 if clearing a non-container value.",
```

PR #309, opened 2025-06-06
Adds docs for Valkey LDAP module

rjd15372:
Documentation for using the LDAP module https://github.com/valkey-io/valkey-ldap

### DIFF
--- a/topics/ldap.md
+++ b/topics/ldap.md
@@ -33,7 +33,7 @@ uid: ben
 
 In this case, the DN is `uid=ben,ou=engineering,dc=valkey,dc=io`, and the username is `ben`. To use `bind` mode, set the prefix to `uid=` and the suffix to `,ou=engineering,dc=valkey,dc=io`.
 
-You can configure these values using the `ldap.bind_prefix` and `ldap.bind_suffix` options.
+You can configure these values using the `ldap.bind_dn_prefix` and `ldap.bind_dn_suffix` options.
 
 ## Search+Bind Authentication
 
@@ -127,17 +127,17 @@ The module extends the output of the `INFO` command by adding a new section call
 
 The `ldap_status` section includes a dictionary entry for each server, with the following possible fields:
 
-- `url`: The server URL as specified in the `ldap.servers` configuration.
+- `host`: The server hostname extracted from the URL specified in the `ldap.servers` configuration.
 - `status`: Indicates the server's health. Possible values are `healthy` (the server is reachable and responds to LDAP operations) or `unhealthy` (the server cannot process LDAP operations).
-- `ping_time`: The round-trip time (RTT) for a simple LDAP operation, in milliseconds. This field is shown only for `healthy` servers.
+- `ping_time_ms`: The round-trip time (RTT) for a simple LDAP operation, in milliseconds. This field is shown only for `healthy` servers.
 - `error`: A description of the error that caused the server to be marked as `unhealthy`. This field is shown only for `unhealthy` servers.
 
 Example output from the `INFO` command:
 
 ```
 # ldap_status
-ldap_server_0:url=ldap://<hostname>,status=unhealthy,error=<some error message>
-ldap_server_1:url=ldaps://<hostname>,status=healthy,ping_time(ms)=1.645
+ldap_server_0:host=<hostname>,status=unhealthy,error=<some error message>
+ldap_server_1:host=<hostname>,status=healthy,ping_time_ms=1.645
 ```
 
 ## Advanced Configuration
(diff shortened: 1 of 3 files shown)


nastena1606 (approved):
Lgtm

topics/ldap.md
martinrvisser, line 1: "Valkey versions 7.2.x and above" should that include module API V1?

martinrvisser, line 1: "and the suffix to ,ou=engineering,dc=valkey,dc=io. " maybe emphasise the need for the comma since it's a suffix (unless you have validation on that

martinrvisser, line 1: "using the ACL SETUSER command" or through entries in valkey.conf or an ACL file ( doc link?)

martinrvisser, line 1: "The configuration option ldap.server "  Should the parameter be "servers"?

martinrvisser: "The authentication mode bind vs search+bind is configured using the ldap.auth_mode option." point out somewhere that all these parameters are set using CONFIG SET

martinrvisser, line 1: "Advanced Options" add some more description about when to adjust these parameters?

rjd15372: I don't think it's needed because API V1 is the version in 7.2.x and if the module API version gets bumped we will also support it.

topics/ldap.md
```diff
@@ -0,0 +1,212 @@
...
+- `status`: Indicates the server's health. Possible values are `healthy` (the server is reachable and responds to LDAP operations) or `unhealthy` (the server cannot process LDAP operations).
+- `ping_time`: The round-trip time (RTT) for a simple LDAP operation, in milliseconds. This field is shown only for `healthy` servers.
+- `error`: A description of the error that caused the server to be marked as `unhealthy`. This field is shown only for `unhealthy` servers.
+
+Example output from the `INFO` command:
+
+```
+# ldap_status
+ldap_server_0:url=ldap://<hostname>,status=unhealthy,error=<some error message>
+ldap_server_1:url=ldaps://<hostname>,status=healthy,ping_time(ms)=1.645
```
madolson, line 140:
```suggestion
ldap_server_1:url=ldaps://<hostname>,status=healthy,ping_time_ms=1.645
```
Typically info fields don't have parenthesis

topics/ldap.md
```diff
@@ -0,0 +1,212 @@
...
+
+To ensure that `bob` cannot log in using password-based authentication, you can create the Valkey user without specifying a password.
+
+To create a user without a password, simply omit the password when running the `ACL SETUSER` command and define the appropriate ACL rules. For example:
+
+```
+ACL SETUSER bob on +@hash
+```
+
+After creating the user `bob` in Valkey, authentication will only be possible through the LDAP module.
```
madolson, line 94: It would be nice if there was some way to restrict a user to just a given module. Not something we have today, but something we might want to consider in the future.

topics/ldap.md
```diff
@@ -0,0 +1,212 @@
...
+| `ldap.search_attribute`  | string                       | `"uid"`           | LDAP attribute to match against the username provided in the `AUTH` command.                     |
+| `ldap.search_scope`      | Enum(`base`, `one`, `sub`)   | `sub`             | Scope of the LDAP search.                                                                        |
+| `ldap.search_dn_attribute`| string                      | `"entryDN"`       | Attribute containing the DN of the user entry.                                                   |
+
+### Advanced Options
+
+| Config Name                  | Type   | Default | Description                                                                                      |
+|------------------------------|--------|---------|--------------------------------------------------------------------------------------------------|
+| `ldap.connection_pool_size`  | number | `2`     | Number of connections in each LDAP server's connection pool.                                      |
+| `ldap.failure_detector_interval` | number | `1`  | Interval (in seconds) between each failure detector check.                                        |
```
madolson, line 209: Does this include exponential backoff? If you have say, 1000 nodes, and there is a temporary disconnect. You might have a 1000 nodes all trying to re-establish connection to the LDAP system.

madolson (commented):
Content looks really good to me, mostly some minor comments.

topics/ldap.md
madolson: Sure, my concern is that if you have a lot of Valkey servers that are connected to the same pool of LDAP servers. When they disconnect and reconnect, their could be thundering herd. I think having exponential backoff is a good idea to do by default.

rjd15372: That's a good point. I'm going to open an issue in the valkey-ldap project to implement the exponential backoff.

rjd15372: https://github.com/valkey-io/valkey-ldap/issues/39

madolson (approved):

PR #310, opened 2025-06-09
Creating a getting started guide for bloom

zackcam:
~~Adding a getting started guide for bloom filters~~

~~This will include the following:~~
~~- Manual Building / installation (Valkey Server start up +  Module Load)~~
~~- Valkey Extension Docker based startup~~
~~- Examples of executing bloom commands~~
~~- Snippet of executing bloom commands using a python script with valkey-py~~
~~- Snippet / examples of using bloom configurations~~

~~We will need to decide where we will show getting started guides as currently they are quite hidden in documentation. But for now I have put under tutorials and FAQ. If we want a more explicit/obvious place let me know and i will look at making that change~~

After discussion we have decided to move the getting started information from a separate page.
The sections we will now include is a tutorial, and a better explanation of the configs for bloom filter. We will save the extension and python example for separate pages that are yet to be created. When they are created I can add the sections from before to them.


zackcam:
The two spell check errors I think aren't correct: reproducibility is a word and spelled correctly, while pre-built should be hyphenated. If people agree I will add them to the excluded words list

topics/bloomfilters.md
```diff
@@ -58,6 +58,97 @@ In this username example, we can use a Bloom filter to track every username that
...
+Once valkey-bloom is built, you can run the Valkey server with the module loaded in two different ways, on server start up:
+```bash
+./valkey-server --loadmodule ./target/release/libvalkey_bloom.so
+```
+You can also load the valkey bloom module on an already running server by running the following command in [valkey-cli](cli.md):
+```bash
+127.0.0.1:6379> MODULE LOAD /path/to/libvalkey_bloom.so
+```
+
+### Usage example
```
KarthikSubbarao, line 81:
```suggestion
### CLI Usage example
```

Also, can we have a small section at the end which includes valkey-py code for bloom?

topics/bloomfilters.md
```diff
@@ -58,6 +58,97 @@ In this username example, we can use a Bloom filter to track every username that
 * If no, the user is created and the username is added to the Bloom filter.
 * If yes, the app can decide to either check the main database or reject the username.
 
+## Tutorial
```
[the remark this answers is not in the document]
zackcam: I tried to follow how sets had it because they have it on top of performance and limits. But also wanted to have it after examples because that way the tutorial example makes more sense and provides context on that

topics/bloomfilters.md
zackcam: I think there is meant to be a specific valkey-py page at some point where we can post the code for that. I have the code ready for that. I'd be happy to add here but there was already some concerns on the length of this

zackcam:
Just a note as I am linking to the extensions page, that pr will need to be finished first

topics/bloomfilters.md
```diff
@@ -58,6 +58,88 @@ In this username example, we can use a Bloom filter to track every username that
...
+
+### Setup
+
+1. **Install Valkey**: Follow the [installation guides](installation.md) to set up Valkey on your system.
```
KarthikSubbarao, line 65: Why is this needed? it is not needed if someone is using Docker, right?

If you agree, lets remove this

topics/bloomfilters.md
```diff
@@ -58,6 +58,84 @@ In this username example, we can use a Bloom filter to track every username that
...
+14) "0.5"
+15) Max scaled capacity
+16) (integer) 26214300
+```
+* If you know the exact number of users who will sign up, or if want to limit the number of items regardless, you can create a non scaling bloom filter. This will create a filter that has capacity for only 1000 items with a 1 in 10,000 chance of a false positive.
+```bash
+127.0.0.1:6379> BF.RESERVE limited_users 0.00001 1000 NONSCALING
+OK
+```
+* If you anticipate once initial users have signed up after a while you will get a large increase in users you can set a custom expansion rate.
```
KarthikSubbarao, line 133:
```suggestion
* If you anticipate that, once initial users have signed up, you will get a large increase in users  after a while, you can set a custom expansion rate.
```

topics/bloomfilters.md
```diff
@@ -58,6 +58,84 @@ In this username example, we can use a Bloom filter to track every username that
...
+ 9) Error rate
+10) "0.001"
+11) Expansion rate
+12) (integer) 2
+13) Tightening ratio
+14) "0.5"
+15) Max scaled capacity
+16) (integer) 26214300
+```
+* If you know the exact number of users who will sign up, or if want to limit the number of items regardless, you can create a non scaling bloom filter. This will create a filter that has capacity for only 1000 items with a 1 in 10,000 chance of a false positive.
```
KarthikSubbarao, line 128: Do you mean 1 in 100,000?

The example below uses 0.00001. We can correct either one

topics/bloomfilters.md
```diff
@@ -58,6 +58,84 @@ In this username example, we can use a Bloom filter to track every username that
...
+## Tutorial
+
+**There are two ways to run valkey-bloom:**
+   - **Docker**: Follow the steps here for using the [valkey bundle](valkey-bundle.md)
+   - **Build from source**: Follow the [build instructions](https://github.com/valkey-io/valkey-bloom/blob/unstable/README.md#build-instructions)
+
+
+### Usage example with CLI
+
+* Create a bloom filter of taken usernames with a 1 in 1000 chance of false positives and 10000 initial capacity.
```
KarthikSubbarao, line 70:
```suggestion
* Create a bloom filter to track the usernames which are taken with a 1 in 1000 chance of a false positive and 10000 as the initial capacity.
```

topics/bloomfilters.md
```diff
@@ -134,6 +212,47 @@ Example of default bloom filter information:
...
+
+Example usage of increasing the limit:
+```bash
+127.0.0.1:6379> CONFIG SET bf.bloom-memory-usage-limit 268435456
+```
+
+Having a limit on the max memory per bloom object prevents them from growing unbounded, thus maintaining server performance during serialization.
+If your use case requires bloom filters beyond this limit, you can increase this configuration value.
+
+However, be aware it can impact overall system performance and memory availability for other operations.
```
KarthikSubbarao, line 252:
```suggestion
If your use case requires bloom filters with capacity beyond what this limit supports, you can increase this configuration value. However, be aware it can impact overall system performance and memory availability for other operations.
```

topics/bloomfilters.md
```diff
@@ -134,6 +212,47 @@ Example of default bloom filter information:
...
+
+When a bloom filter reaches this memory limit, any operation that would cause it to exceed the limit will fail with an error message indicating that the memory limit would be exceeded.
```
KarthikSubbarao, line 254: Do we want an example below?

topics/bloomfilters.md
```diff
@@ -134,6 +212,47 @@ Example of default bloom filter information:
 16) (integer) 26214300
 ```
 
+### When to adjust default configurations
+
+Adjusting the default configurations below can be used to update the behavior of the default bloom object created and are beneficial in different scenarios:
+
+- **Higher capacity (bf.bloom-capacity)**: Increase this value when you expect to store a larger number of items in all your new bloom filter objects. It increases the overall capacity for non scaling filters. For scaling filters, it increases the starting/initial capacity which reduces the need for additional scaling operations which can improve performance.
```
KarthikSubbarao, line 219:
```suggestion
- **Higher capacity (bf.bloom-capacity)**: Increase this value when you expect to store a larger number of items in all your new bloom filter objects. It increases the overall capacity for non scaling filters. For scaling filters, it increases the starting/initial capacity which potentially reduces the need for additional scaling operations which can improve performance.
```

topics/bloomfilters.md
```diff
@@ -134,6 +212,47 @@ Example of default bloom filter information:
...
+
+- **Lower false positive rate (bf.bloom-fp-rate)**: Decrease this value when correctness is critical for your application and you want to reduce the number of false positives in all your new bloom filter objects. For example, in fraud detection or security applications where false positives are costly.
+
+- **Higher expansion rate (bf.bloom-expansion)**: Increase this value when you want faster growth of bloom filters that will scale across all your new scaling bloom filter objects. This reduces the number of scaling operations but uses more memory quicker.
+
+- **Lower tightening ratio (bf.bloom-tightening-ratio)**: Lower this when you want to maintain a more consistent false positive rate across multiple scaling operations for all your new bloom filter objects. (Not advisable to change this default, as it is already set to a strict value of 0.5)
+
+- **Fixed seed (bf.bloom-use-random-seed)**: Set to false only when you want the new bloom filter object creations to use a fixed seed for deterministic occurrence of false positives during item add/exists operations. It can be used  for testing or reproducibility.
```
KarthikSubbarao, line 227:
```suggestion
- **Fixed seed (bf.bloom-use-random-seed)**: Set to false only when you want the new bloom filter object creations to use a fixed seed for deterministic occurrence of false positives during item add/exists operations. It can be used for testing or reproducibility.
```

topics/bloomfilters.md
zackcam: Added

KarthikSubbarao (approved):

topics/bloomfilters.md
```diff
@@ -201,6 +325,13 @@ There are two notable validations bloom filters faces.
 
     The memory usage limit per bloom filter by default is defined by the `BF.BLOOM-MEMORY-USAGE-LIMIT` module configuration which has a default value of 128 MB. If a command results in a creation / scale out causing the overall memory usage to exceed this limit, the command is rejected. This config is modifiable and can be increased as needed.
 
+The `BF.INFO` command's `SIZE` field can be used to find out the current size of a bloom filter.
+
+```bash
+172.31.45.25:6379> BF.INFO validate_scale_valid SIZE
```
KarthikSubbarao, line 331: Should we swap this with `127.0.0.1`? :D

KarthikSubbarao:
This PR looks good to me and I approved it

PR #313, opened 2025-06-12
Bump to the latest patch version: 8.1.2

hpatro:
Seems like we added at this step at a later point to the release document https://github.com/valkey-io/valkey/wiki/Releasing-a-new-Valkey-version#new-patch-release or missed it for 8.1.1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 # Version info
-VERSION ?= 8.1.0
-DATE ?= 2025-03-31
+VERSION ?= 8.1.2
+DATE ?= 2025-06-11
 
 # Path to the code repo.
 VALKEY_ROOT ?= ../valkey


zuiderkwast (commented):
This is only used for the footer of man pages. It's not that important if it says 8.1.0 because the patch releases shouldn't add any features anyway.

Did we already set a tag 8.1.2 in the doc repo? The doc repo main branch  already contains a lot of 9.0 features that don't exist in 8.1.2. We can just set the doc tag 8.1.2 to point to the same commit as 8.1.0, if you didn't set the tag yet.

For the patch releases in the older tracks, I have just set the tag 8.0.3 to point to the same commit as 8.0.0.

We do these tags mainly for packagers (fedora) so they can get reproducible builds with docs.

hpatro:
Alright thanks for the context. I've not pushed the tags. Was waiting for your review. Sounds like we should leave it as is. 

Maybe we should update the patch release wiki to avoid this step as we can get in next version related documentation tagged incorrectly, which will be quite confusing for end users.

PR #314, opened 2025-06-17
Valkey Bundle Getting Started Guide

Nikhil-Manglore:
This is the getting started guide for Valkey Bundle. The commands will have to be updated once the name is changed from Valkey Extensions to Valkey Bundle.

### DIFF
--- a/topics/valkey-bundle.md
+++ b/topics/valkey-bundle.md
@@ -1,0 +1,173 @@
+## Valkey Bundle Quick Start Guide
+
+# What is Valkey Bundle
+
+Valkey Bundle is a pre-packaged, containerized version of Valkey that comes preloaded with a collection of supported modules. These modules enable advanced data structures and search capabilities to extend Valkey’s core functionality. The bundle is designed to help developers quickly get started with powerful Valkey features without needing to manually install or configure anything. Some of the modules included in the bundle are:
+
+1. Valkey JSON - Allows users to natively store, query, and modify JSON data structures using the JSONPath query language.
+2. Valkey Bloom - Provides space-efficient probabilistic data structures, such as Bloom filters, for adding elements and checking whether they exist in a set.
+3. Valkey Search - Enables the creation of indexes and similarity searches through the use of full-text and complex filters. 
+4. Valkey LDAP - Handles user authentication against LDAP based identity providers.  
+
+# Quick Start to Using the Bundle
+
+The fastest way to start using Valkey Bundle is by downloading the official image through [Docker Hub](https://hub.docker.com/r/valkey/valkey-extensions/). The following steps will guide you through launching and interacting with your first instance!
+
+1. **Pull the image to get the latest public release**
+
+    ```bash
+    docker pull valkey/valkey-extensions
+    ```
+
+    This command downloads the most recent stable image of the Valkey Bundle, which includes the Valkey server along with the preloaded modules.
+
+    The Valkey Bundle image also supports multiple tags, allowing you to control the specific version and operating system base. This allows for more control over the environment, whether you’re aiming for a reproducible build (using a version like 8.1-bookworm) or a minimal footprint (alpine variant).
+
+    Check out the [Valkey Bundle Docker Hub Tag](https://hub.docker.com/r/valkey/valkey-extensions/tags) section to view all available tags and example pull commands. 
+
+2. **Start the Valkey server inside a container**
+
+    ```bash
+    docker run --name my-valkey-extensions -d -p 6379:6379 valkey/valkey-extensions
+    ```
+
+    This command starts the Valkey server in a background container named my-valkey-extensions and maps Valkey’s default port 6379 to your local machine for external access. By default, it uses the latest available image. To run a specific version or variant, append the desired tag to the image name. For example:
+
+    ```bash
+    docker run --name my-valkey-extensions -d -p 6379:6379 valkey/valkey-extensions:8.1-bookworm
+    ```
+ 
+3. **Connect to the server using valkey-cli**
+    
+    To interact with the server, use the Valkey command-line interface (valkey-cli). The example below runs the CLI from a separate container on the same Docker network.
+     
+    ```bash 
+    docker run -it --network some-network --rm valkey/valkey-extensions valkey-cli -h my-valkey-extensions
+    ```
+    
+    This approach is ideal when you want to isolate your CLI session or connect from other containers. However, it does require both containers to be connected to the same Docker network.
+
+    Alternatively, you can run the CLI directly inside the server container: 
+
+    ```bash
+    docker exec -it my-valkey-extensions valkey-cli
+    ```
+
+    This runs the valkey-cli directly inside the server container and connects to the server on localhost. This approach requires no extra setup but only works when the server container is running locally and accessible from your host environment.
+
+4. **Try some commands**
+
+    ```
+    # Check Loaded Modules
+
+    MODULE LIST
+
+    # Use the Valkey JSON Module
+    JSON.SET test $ ‘{"hello": "world"}’
+    JSON.GET test
+
+    # Use the Valkey Bloom Module
+    BF.RESERVE test_bloom 0.01 1000
+    BF.ADD test_bloom "item1"
+    BF.EXISTS test_bloom "item1"
+
+    # Use the Valkey Search Module
+    FT.CREATE index ON HASH PREFIX 1 doc: SCHEMA name TEXT
+    HSET doc:1 name "valkey bundle"
+    FT.SEARCH index "bundle"
+    ```
+
+    Make sure to check out the full list of commands for all the bundle components:
+
+    1. [Valkey Commands](https://valkey.io/commands/)
+    2. [Valkey JSON Commands](https://valkey.io/commands/#json)
+    3. [Valkey Bloom Commands](https://valkey.io/commands/#bloom)
+    4. [Valkey Search Commands](https://valkey.io/commands/#search)
+
+Valkey Bundle supports more advanced setup options too including:
+
+1. **Persistent storage**
+    
+    Persistent storage allows you to save data snapshots locally. The command below is an example of how you can save a snapshot every 60 seconds if at least one write occurred. 
+    
+    ```
+    bash docker run --name my-valkey-extensions -d valkey/valkey-extensions valkey-server --save 60 1 --loglevel warning
+    ```
+    
+2. **Custom Flags with Environment Variable**
+    
+    This allows you to pass additional Valkey flags at runtime using the VALKEY_EXTRA_FLAGS environment variable. This is a flexible way for customizing behavior without needing to modify the existing image or use a custom configuration file. 
+    
+    ```
+    bash docker run --env VALKEY_EXTRA_FLAGS='--save 60 1 --loglevel warning' valkey/valkey-extensions
+    ```
+    
+3. **Use a Custom Configuration File**
+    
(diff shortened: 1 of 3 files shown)


topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+4. Valkey LDAP - Handles user authentication against LDAP based identity providers.  
+
+# Quick Start to Using the Bundle
+
+The fastest way to start using Valkey Bundle is by downloading the official image through [Docker Hub](https://hub.docker.com/r/valkey/valkey-extensions/). The following steps will guide you through launching and interacting with your first instance!
+
+1. **Pull the image to get the latest public release**
+
+    ```bash
+    docker pull valkey/valkey-extensions
```
sarthakaggarwal97, line 19: this will change to valkey-bundle right?

sarthakaggarwal97:
This still pulls up the `valkey-extensions` from docker. Do we know the plan to update the name in docker as well to `valkey-bundle`. @roshkhatri might know more about this.

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+    
+    This allows you to pass additional Valkey flags at runtime using the VALKEY_EXTRA_FLAGS environment variable. This is a flexible way for customizing behavior without needing to modify the existing image or use a custom configuration file. 
+    
+    ```
+    bash docker run --env VALKEY_EXTRA_FLAGS='--save 60 1 --loglevel warning' valkey/valkey-extensions
+    ```
+    
+3. **Use a Custom Configuration File**
+    
+    If you need full control over your Valkey settings, you can create a customer configuration file that will be used inside the container. This allows you to override all default settings through the use of your own valkey.conf file.
```
madolson, line 107: This step is done without context. You are assuming there is a directory (/myvalkey/conf) that you are mounting in the container. This folder must have a valkey.conf file. You should explain what's happening here.

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+
+    Check out the [Valkey Bundle Docker Hub Tag](https://hub.docker.com/r/valkey/valkey-extensions/tags) section to view all available tags and example pull commands. 
+
+2. **Start the Valkey server inside a container**
+
+    ```bash
+    docker run --name my-valkey-extensions -d -p 6379:6379 valkey/valkey-extensions
+    ```
+
+    This command starts the Valkey server in a background container named my-valkey-extensions and maps Valkey’s default port 6379 to your local machine for external access. By default, it uses the latest available image. To run a specific version or variant, append the desired tag to the image name. For example:
```
madolson, line 34:
```suggestion
    This command starts the Valkey server in a container named my-valkey-extensions and maps Valkey’s default port 6379 to your local machine for external access. By default, it uses the latest available image. To run a specific version or variant, append the desired tag to the image name. For example:
```

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+    docker run --name my-valkey-extensions -d -p 6379:6379 valkey/valkey-extensions:8.1-bookworm
+    ```
+ 
+3. **Connect to the server using valkey-cli**
+    
+    To interact with the server, use the Valkey command-line interface (valkey-cli). The example below runs the CLI from a separate container on the same Docker network.
+     
+    ```bash 
+    docker run -it --network some-network --rm valkey/valkey-extensions valkey-cli -h my-valkey-extensions
+    ```
```
madolson, line 46: I'm not really sure why you would do this. I think someone should either have the cli installed on their local machine or they should run the cli from inside the environment. There is really no good reason to run it from another container. I would remove this specifically because it won't work as written since there is no "some-network".

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+    
+    This approach is ideal when you want to isolate your CLI session or connect from other containers. However, it does require both containers to be connected to the same Docker network.
+
+    Alternatively, you can run the CLI directly inside the server container: 
+
+    ```bash
+    docker exec -it my-valkey-extensions valkey-cli
+    ```
+
+    This runs the valkey-cli directly inside the server container and connects to the server on localhost. This approach requires no extra setup but only works when the server container is running locally and accessible from your host environment.
```
madolson, line 56:
```suggestion
    This runs the valkey-cli directly inside the server container and connects to the server on localhost. This approach requires no extra setup but only works when the server container is running locally and accessible from your host environment.
```

As written you are right, but the docker daemon can be configured to be accessible on another host.

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+
+4. **Try some commands**
+
+    ```
+    # Check Loaded Modules
+
+    MODULE LIST
+
+    # Use the Valkey JSON Module
+    JSON.SET test $ ‘{"hello": "world"}’
```
madolson, line 66:
```suggestion
    JSON.SET test $ '{"hello": "world"}'
```

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+    JSON.GET test
+
+    # Use the Valkey Bloom Module
+    BF.RESERVE test_bloom 0.01 1000
+    BF.ADD test_bloom "item1"
+    BF.EXISTS test_bloom "item1"
+
+    # Use the Valkey Search Module
+    FT.CREATE index ON HASH PREFIX 1 doc: SCHEMA name TEXT
```
madolson, line 75: This isn't create a hash. I also think it has to be a VECTOR schema for now? I would borrow the introduction from https://valkey.io/commands/ft.search/

topics/valkey-bundle.md
madolson, line 107:
```suggestion
    If you need full control over your Valkey settings, you can create a custom configuration file that will be used inside the container. This allows you to override all default settings through the use of your own valkey.conf file.
```

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+    
+    Possible Causes: 
+
+    - The container name is already in use.
+    
+    - The host port is already taken by another process.
+    
+    Solutions:
+
+    - Run docker ps -a to check for existing containers with the same name. If the name already exists, create the Docker container using a different name.
```
madolson, line 128: When you do the docker run command, you get very explicit errors in this case. Either the name is taken or the bind address is already in use.

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+    
+    Possible Causes: 
+
+    - The CLI container and server container aren’t on the same network.
+    
+    - The server isn’t running or hasn’t finished starting up.
+    
+    Solutions:
+    
+    - Create and use a user-defined Docker network or run the CLI from inside the container.
```
madolson, line 142: I have *never* setup a docker network. Im sure some folks have, but it's not very common. I would avoid the usage of the docker networks.

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+    
+3. **Modules aren’t loaded**
+    
+    Possible Causes: 
+
+    - You might be using the base valkey image instead of valkey/valkey-bundle.
+    
+    - The image didn’t load the modules due to an error.
```
madolson, line 150: You don't provide any mitigation for this, and I don't really think it's a real trouble shouting step.

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
...
+    Solutions:
+
+    - Run the command MODULE LIST inside the CLI to confirm loaded modules.
+    
+    - Make sure you’re using the Valkey Bundle image and not the Valkey image by mistake.
+    ```
+    bash docker run --name my-valkey-extensions -d valkey/valkey-extensions
+    ```
+    
+# Next Steps
```
madolson, line 161:
```suggestion
## Next Steps
```

topics/valkey-bundle.md
```diff
@@ -0,0 +1,173 @@
+## Valkey Bundle Quick Start Guide
+
+# What is Valkey Bundle
+
+Valkey Bundle is a pre-packaged, containerized version of Valkey that comes preloaded with a collection of supported modules. These modules enable advanced data structures and search capabilities to extend Valkey’s core functionality. The bundle is designed to help developers quickly get started with powerful Valkey features without needing to manually install or configure anything. Some of the modules included in the bundle are:
+
+1. Valkey JSON - Allows users to natively store, query, and modify JSON data structures using the JSONPath query language.
+2. Valkey Bloom - Provides space-efficient probabilistic data structures, such as Bloom filters, for adding elements and checking whether they exist in a set.
+3. Valkey Search - Enables the creation of indexes and similarity searches through the use of full-text and complex filters. 
+4. Valkey LDAP - Handles user authentication against LDAP based identity providers.
```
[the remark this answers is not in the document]
madolson: We also don't support full-test search.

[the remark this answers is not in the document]
Nikhil-Manglore: I just noticed you said to remove the call to action at the bottom, so i'll move the links up here and remove that section

topics/valkey-bundle.md
Nikhil-Manglore: I think I will remove the next steps section entirely and move the links for documentation up to the "What is Valkey Bundle" section.

topics/valkey-bundle.md
Nikhil-Manglore: I think it might be better to remove the trouble shooting section entirely. I feel that all the problems I listed aren't really issues that users might face.

madolson (commented):
There is a lot of overlap between this and just running the normal container. Should we have a single doc for how to run with a container? (Most of this content is still fine, but just wondering out allowed if we should generalize this)

topics/valkey-bundle.md
```diff
@@ -0,0 +1,130 @@
...
+    This allows you to override the default settings using your own valkey.conf file.
+
+    First, ensure you have a valkey.conf file located in a local directory, such as /myvalkey/conf/valkey.conf. 
+    The valkey.conf file can include any standard Valkey configuration directives such as memory limits, save intervals, logging levels, and more. The file has a very simple format:
+
+    ```bash
+    keyword arg1 arg2 ... argN
+    ``` 
+
+    Check out the [Configuration Documentation](https://valkey.io/topics/valkey.conf/) to learn more.
```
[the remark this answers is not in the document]
Nikhil-Manglore: That makes sense. I'll include this sample so they don't have to search through the blog for this example.

Nikhil-Manglore:
> There is a lot of overlap between this and just running the normal container. Should we have a single doc for how to run with a container? (Most of this content is still fine, but just wondering out allowed if we should generalize this)

I agree that there is overlap with how to run a container. I think maybe we can rename this guide to something along the lines of "getting started with valkey containers". Then we start out with explaining starting docker containers for valkey. Then we can have a separate section after where we explain the extra bundle stuff. 

On the other hand, I know that the getting started guide for Valkey already explains setting up the docker container as well. So maybe we could just keep this as is and consolidate everything about getting started with the bundle from scratch in this guide. I don't really have a preference but I know that the valkey bundle blog also explains setting up the container. @rlunar What are your thoughts?

rlunar:
> t

I know there's overlap but I like every Getting Started page to stand on its own; as an end user I don't want to jump around to find the bits and pieces to make things work; as a new user I would like as much info as possible, as an experienced user I'll just skim through everything I know and move on quickly to the new parts.

PR #332, opened 2025-07-09
Add command hyperlinks for server management command files

milliehartnt123:
The following files were modified. Some of the files have two sets of commits, with one set being to add the tick marks back to the hyperlink for formatting.
* acl.md
* acl-getuser.md
* acl-list.md
* acl-log.md
* acl-setuser.md
* acl-whoami.md
* bgrewriteaof.md
* bgsave.md
* command.md
* getkeys.md
* getkeysandflags.md
* command-info.md
* commandlog.md
* commandlog-get.md
* commandlog-len.md
* config.md
* config-get.md
* config-resetsat.md
* config-rewrite.md
* config-set.md
* failover.md
* info.md
* lastsave.md
* latency.md
* latency-graph.md
* latency-histogram.md
* latency-latest.md
* memory.md
* memory-stats.md
* module.md
* module-loadex.md
* module-unload.md
* monitor.md
* save.md
* shutdown.md
* sync.md

### DIFF
--- a/commands/commandlog-len.md
+++ b/commands/commandlog-len.md
@@ -6,4 +6,4 @@ The maximum number of entries in the different command log is governed by the `c
 
 Once the command log reaches its maximal size, the oldest entry is removed whenever a new entry is created.
 
-The command log can be cleared with the [`COMMANDLOG RESET`](connandlog-reset.md) command.
+The command log can be cleared with the [`COMMANDLOG RESET`](commandlog-reset.md) command.


zuiderkwast (commented):
Great @milliehartnt123, thank you!

@stockholmux we have been talking about making these links automatic. I think explicit links is pretty good. In this way we can be sure we don't link wrong for some words that look like commands but are not. WDYT?

stockholmux:
@zuiderkwast yeah, I would be for linking explicitly instead of some script or regex that might be confusingly wrong.

Is this something we'd manually check for on PR review? E.g. someone submits docs is that going to be noted as feedback?

zuiderkwast:
> Is this something we'd manually check for on PR review? E.g. someone submits docs is that going to be noted as feedback?

Yeah, I suppose we will ask contributors to add links to commands when missing. It will be more natural when it is seen everywhere in existing docs. Should we have some automatic checker?

IMO we don't need to link all occurrences, but the first occurrence on each page would be good to link, at least.

PR #376, opened 2025-10-24
Add Valkey 9.0 valkey.conf

murphyjacob4:


madolson (approved):

murphyjacob4:
> Kind of weird we have self documented redis.conf files

Yeah, not sure why folks would go to Valkey's page to read those

PR #413, opened 2026-02-19
Add Spring Data Valkey to GLIDE clients

jeremyprime:
## Summary

Add the Spring Data Valkey client to the GLIDE clients list.  Links to the Spring Data Valkey [GitHub repo](https://github.com/valkey-io/spring-data-valkey) and the [docs site](https://spring.valkey.io/).


jeremyprime:
> Is Spring Data a client? It feels more like a library or an integration than a client, which is a gap of our current documentatino.

Agreed, it is more of a framework (Spring) on top of a client (GLIDE).  But I didn't think it should get a top-level link [here](https://valkey.io/docs/) or be buried under the [topics](https://valkey.io/topics/), so putting it under the clients page seems to be the best fit for now.

PR #419, opened 2026-03-07
Add note on RDB version import error

Sp1l:
While migrating Redis 8.6.1 to Valkey 9.0.3 on FreeBSD 15.0-p4 amd64, I ran into the following error

```log
6431:M 07 Mar 2026 09:37:43.232 # Can't handle RDB format version 12
````

Document the simple fix.

### DIFF
--- a/topics/migration.md
+++ b/topics/migration.md
@@ -106,6 +106,8 @@ To perform a physical migration:
 8. Stop the Redis server.
 9. Start Valkey:
    > NOTE: If you enabled AOF in your Valkey configuration, disable it on the first start. Otherwise, the copied RDB file will not be imported into Valkey.
+   
+   > NOTE: Set `rdb-version-check relaxed` for the first start to prevent "RDB format version" errors.
 
    Run the following command:
 

### CODE REVIEW

File: topics/migration.md
Comment (zuiderkwast) [inline topics/migration.md:35]: This table and the whole page needs some updates. Currently, it only describes a plain upgrade from Redis 7.2 before the two projects started to diverge.

This page doesn't describe the relaxed RDB version check. It should, and it the limitations need to be clearly explained, i.e. that it no new keys like Redis hash-field expiration can be present, and that reading the RDB file may or may not work depending on what Redis added in their RDB files. This matrix should describe the versions we have tested and confirmed.

File: topics/migration.md [hunk 1]
```diff
@@ -106,6 +106,8 @@ To perform a physical migration:
 8. Stop the Redis server.
 9. Start Valkey:
    > NOTE: If you enabled AOF in your Valkey configuration, disable it on the first start. Otherwise, the copied RDB file will not be imported into Valkey.
+   
+   > NOTE: Set `rdb-version-check relaxed` for the first start to prevent "RDB format version" errors.
```
zuiderkwast, line 110: Loading an RDB with the relaxed check can easily fail when an unknown RDB opcode is encountered, and then this page is not of much help. We need more elaborate documentation.

zuiderkwast (commented):
Thanks! This page as a whole currently doesn't describe the challenges with migrating from Redis > 7.2. The page as a whole needs more love. The simple note can help some but it can also give false hopes in the situations where it doesn't work.

zuiderkwast:
@nastena1606

PR #422, opened 2026-03-16
Search Module - Misc Updates + Metrics + Configurations 

KarthikSubbarao:
Search Module - Misc Updates + Metrics + Configurations


commands/ft.aggregate.md
```diff
@@ -92,3 +92,5 @@ The following reducer functions are available. The reducer functions that take a
 | MAX 1 <expression>            | The largest numerical values of the expression.                                                                                    |
 | AVG 1 <expression>            | The numerical average of the values of the expression.                                                                             |
 | STDDEV 1 <expression>         | The standard deviation the values of the expression.                                                                               |
+
+Note: Text Indexing and Search support was added in valkey-search version 1.2.
```
madolson: This should be part of the history information, not just put at the bottom. Is there a way we can include that properly?

topics/search-telemetry.md
```diff
@@ -0,0 +1,126 @@
+---
+title: "Valkey Search - Metrics and Configurables"
```
madolson:
```suggestion
title: "Valkey Search - Metrics and configurations"
```

topics/search-monitoring.md
madolson, line 1: Why is this called telemetry? I think people will be confused if we call this telemetry since that is more associated with traces.

madolson (commented):
Some fast high level things. Doing a deeper review.

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| search.coordinator-query-timeout-secs | Number |   120   | Controls the gRPC deadline timeout (in seconds) for distributed coordinator query operations.      |
+| search.max-indexes                   | Number |   1000   | Controls the maximum number of search indexes that can be created in the system                    |
+| search.prefiltering-threshold-ratio  | String |   0.001   | Controls when pre-filtering is used vs inline-filtering for hybrid queries (range 0.0 to 1.0) |
+| search.async-fanout-threshold        | Number |   30   | Controls the threshold for async fanout operations (minimum number of targets to use async) |
+| search.max-nonvector-search-results-fetched | Number |   100000   | Controls the maximum number of results to fetch in background threads before content fetching on non-vector (numeric/tag/text) query paths |
+| search.max-attributes                | Number |   1000   | Controls the max number of attributes per index |
+| search.hide-user-data-from-log       | Boolean |   true   | Controls whether user data is hidden from log output |
```
madolson: Why is this separate from the core config?

topics/search-telemetry.md
```diff
@@ -0,0 +1,126 @@
...
+| hnsw_search_exceptions_count                                   |     hnswlib      |    Count     | Count of exceptions during HNSW vector searches                                                                                                                                   |
+| number_of_attributes                                           |   index_stats    |    Count     | Total count of all attributes across all search indexes                                                                                                                                  |
+| number_of_indexes                                              |   index_stats    |    Count     | Total number of search indexes                                                                                                                                                     |
+| total_active_write_threads                                     |   index_stats    |    Count     | "Number of active writer threads (0 if suspended, otherwise the writer thread pool size)"                                                                                         |
```
madolson: Missing 3 metrics that appear in default `INFO SEARCH` output (verified against `valkey-bundle:unstable`):
- `number_of_active_indexes` (index_stats)
- `number_of_active_indexes_indexing` (index_stats)
- `number_of_active_indexes_running_queries` (index_stats)

topics/search-telemetry.md
```diff
@@ -0,0 +1,126 @@
...
+| total_indexed_documents                                        |   index_stats    |    Count     | Total number of documents indexed across all indexes                                                                                                                              |
+| total_indexing_time                                            |   index_stats    | Milliseconds | Total indexing time                                                                                                                                                               |
```
madolson: Verified: `total_indexing_time` always returns 0 in the running container (source has a TODO). Unit "Milliseconds" is unverifiable — consider noting this is not yet implemented.

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| prefiltering_requests_count                                    |      query       |    Count     | Count of queries using pre-filtering                                                                                                                                              |
+| result_record_dropped_count                                    |      query       |    Count     | Tracks records dropped when ft.search results exceed configured limits                                                                                                            |
+| rdb_load_failure_cnt                                           |       rdb        |    Count     | Number of failed RDB load operations                                                                                                                                              |
+| rdb_load_success_cnt                                           |       rdb        |    Count     | Number of successful RDB load operations                                                                                                                                          |
```
madolson: Verified: actual INFO field name is `query_prefiltering_requests_cnt`, not `prefiltering_requests_count`. This is inconsistent with the other metric names in this table.

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| writer_queue_size                                              |   thread-pool    |    Count     | Current size of the writer thread pool queue                                                                                                                                      |
+| writer_resumed_cnt                                             |   thread-pool    |    Count     | Number of times the writer worker thread pool was resumed                                                                                                                         |
+| writer_suspension_expired_cnt                                  |   thread-pool    |    Count     | Number of times writer suspension expired due to timeout                                                                                                                          |
+| vector_externing_deferred_entry_cnt                            | vector_externing |    Count     | Count of deferred entries in vector externalization                                                                                                                               |
```
madolson: Verified: actual values are decimal fractions (e.g. `0.005`), not 0-100 percentages. "Percent" is ambiguous — clarify whether this is a fraction or a percentage.

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| search.ft-info-rpc-timeout-ms        | Number |   2500   | Timeout in milliseconds for RPC connections used by FT.INFO when CLUSTER OR PRIMARTY options is selected  |
+| search.local-fanout-queue-wait-threshold | Number |   50   | When this value is less than the average read queue wait time (in milliSeconds)  the local node is preferred in a fanout operation. |
+| search.thread-pool-wait-time-samples | Number |   100   | Sample queue size for thread pool wait time tracking   |
+| search.tag-min-prefix-length                  | Number  |   2   |               | Minimum number of characters required before trailing `*` in TAG wildcard queries (length excludes `*`) 
+| search.max-term-expansions           | Number |   200   | Maximum number of words to search in text operations (prefix, suffix, fuzzy) to limit memory usage |
+| search.search-result-buffer-multiplier | String |   1.5   | Multiplier for search result buffer size allocation    |
+| search.drain-mutation-queue-on-save  | Boolean |   false   | Drain the mutation queue before RDB save.           |
```
madolson: Typo: "PRIMARTY" → "PRIMARY"

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| search.query-string-depth            | Number |   1000   | Controls the depth of the query string parsing from the FT.SEARCH cmd        |
+| search.query-string-terms-count      | Number |   1000   | Controls the size of the query string parsing from the FT.SEARCH cmd (number of nodes in predicate tree)                 |
+| search.fuzzy-max-distance            | Number |   3   | Controls the maximum allowed edit distance for fuzzy search queries          |
```
madolson: Broken table formatting — there's an extra `|` creating an empty column between the default value and description. Also, `CONFIG GET "search.tag-min-prefix-length"` returns empty on `valkey-bundle:unstable` — this config may be hidden or not registered in this build.

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| search.max-indexes                   | Number |   1000   | Controls the maximum number of search indexes that can be created in the system                    |
+| search.prefiltering-threshold-ratio  | String |   0.001   | Controls when pre-filtering is used vs inline-filtering for hybrid queries (range 0.0 to 1.0) |
+| search.async-fanout-threshold        | Number |   30   | Controls the threshold for async fanout operations (minimum number of targets to use async) |
+| search.max-nonvector-search-results-fetched | Number |   100000   | Controls the maximum number of results to fetch in background threads before content fetching on non-vector (numeric/tag/text) query paths |
+| search.max-attributes                | Number |   1000   | Controls the max number of attributes per index |
+| search.hide-user-data-from-log       | Boolean |   true   | Controls whether user data is hidden from log output |
+| search.cluster-map-expiration-ms     | Number |   250   | Controls how long (in milliseconds) the coordinator caches the cluster topology map before refreshing it from the Valkey cluster. |
```
madolson: Verified against `valkey-bundle:unstable`: several configs visible via `CONFIG GET "search.*"` are missing from this table:
- `search.info-developer-visible` (default: `no`) — controls dev metric visibility
- `search.drain-mutation-queue-on-load` (default: `yes`)
- `search.rdb-read-v2` (default: `yes`)
- `search.rdb-write-v2` (default: `yes`)
- `search.rdb-validate-on-write` (default: `no`)
- `search.text-rax-target-mutex-pool-size` (default: `256`)
- `search.fanout-data-uniformity` (default: `0`)
- `search.fanout-uniformity-min-index-size` (default: `10000`)

topics/search-telemetry.md
```diff
@@ -0,0 +1,126 @@
...
+| vector_externing_generated_value_cnt                           | vector_externing |    Count     | Count of generated values during vector externalization                                                                                                                           |
+| vector_externing_hash_extern_errors                            | vector_externing |    Count     | Count of errors during hash externalization                                                                                                                                       |
+| vector_externing_lru_promote_cnt                               | vector_externing |    Count     | Number of LRU promotions in vector externalization                                                                                                                                |
+| vector_externing_num_lru_entries                               | vector_externing |    Count     | Number of entries in the vector externalizer LRU cache
+
+## Configurables
```
madolson:
```suggestion
## Configurations
```
Let's be consistent

topics/search-telemetry.md
```diff
@@ -0,0 +1,126 @@
...
+
+The search module uses the Valkey configuration mechanism. Thus each of the named configuration below can be set on the module load command OR via the `CONFIG SET` command.
```
madolson:
```suggestion
The search module uses the Valkey configuration mechanism. Thus each of the named configuration below can be set on the `MODULE LODEX` command OR via the `CONFIG SET` command.
```

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Configurables
+
+The search module uses the Valkey configuration mechanism. Thus each of the named configuration below can be set on the module load command OR via the `CONFIG SET` command.
+
+| Name           |  Type | Default Value | Description                                            |
+| :-------------------------------------------- | :-----: | :-----------: | :-------------------------------------------------------------------------------------------------------------------------------- |
+| search.query-string-bytes            | Number | 10240 | Maximum allowable length of the query string for FT.SEARCH or FT.AGGREGATE commands |
+| search.hnsw-block-size               | Number | 10240 | Number of vectors of space to add to HNSW index size when additional space required. |
+| search.reader-threads                | Number |   #cores   | Reader thread pool size; dynamically resizes pool on modification                     |
```
madolson: What does `#cores` mean? If you have a 12 machine, but reader-threads and writer-threads is set to 12? Is it 12/2?

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| search.writer-threads                | Number |   #cores   | Writer thread pool size; dynamically resizes pool on modification                     |
+| search.utility-threads               | Number |   1   | Utility thread pool size; dynamically resizes pool on modification                    |
+| search.max-worker-suspension-secs    | Number |   60   | Controls how long the worker thread pool quiescing around a fork. Values > 0 are resumption timeouts if a fork runs too long. Values <=0 mean no quiescing.  |
+| search.use-coordinator               | Boolean |   false   | Controls whether this instance uses the coordinator. Generally required for CME; can only be set at startup  |
+| search.skip-rdb-load                 | Boolean |   false   | Skip loading of saved index data from RDB file. May be useful for recovering from a corrupted RDB or index.     |
```
madolson: May be useful or is useful? Will the index be rebuilt or will it just be corrupted?

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| search.skip-corrupted-internal-update-entries | Boolean |  false  | Skip corrupted AOF entries during internal updates. May be useful for recovering from a corrupted AOF file.  |
+| search.log-level                     |  Enum |   from core  | Controls module log level verbosity: "debug", "verbose", "notice" or "warning". Default value is to fetch the log level from the core at startup. |
+| search.enable-partial-results        | Boolean |   true   | Default option for delivering partial results when timeout occurs (uses SOMESHARDS if not explicitly provided)           |
```
madolson: What does `SOMESHARDS` mean in the context of this being a boolean? Is false NO shards? Is false all shards?

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
+---
+title: "Valkey Search - Metrics and Configurables"
+description: Valkey Search Module INFO Fields and Module Configurations
+---
+
+## Metrics - INFO SEARCH
```
madolson, line 6: All the metrics name should be prefixed by search no?

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| coordinator_server_search_index_partition_success_count        |   coordinator    |    Count     | Count of successful server searches on index partitions                                                                                                                           |
+| coordinator_server_search_index_partition_success_latency_usec |   coordinator    | Microseconds | Latency distribution (in microseconds) for successful server partition searches                                                                                                   |
+| coordinator_threads_cpu_time_sec                               |   coordinator    |   Seconds    | Cumulative CPU time consumed by coordinator (gRPC) threads                                                                                                                        |
+| hnsw_add_exceptions_count                                      |     hnswlib      |    Count     | Count of exceptions during HNSW vector additions                                                                                                                                  |
```
madolson: I had this conversation last time we proposed metrics I still don't know what any of these exceptions are or why any end user would care about them.

topics/search-telemetry.md
```diff
@@ -0,0 +1,126 @@
...
+| total_active_write_threads                                     |   index_stats    |    Count     | "Number of active writer threads (0 if suspended, otherwise the writer thread pool size)"                                                                                         |
+| total_indexed_documents                                        |   index_stats    |    Count     | Total number of documents indexed across all indexes                                                                                                                              |
+| total_indexing_time                                            |   index_stats    | Milliseconds | Total indexing time                                                                                                                                                               |
+| background_indexing_status                                     |     indexing     |    String    | "String indicating background indexing status - ""IN_PROGRESS"" or ""NO_ACTIVITY"""                                                                                               |
```
madolson: Why is this in quotes?

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| total_indexing_time                                            |   index_stats    | Milliseconds | Total indexing time                                                                                                                                                               |
+| background_indexing_status                                     |     indexing     |    String    | "String indicating background indexing status - ""IN_PROGRESS"" or ""NO_ACTIVITY"""                                                                                               |
+| flat_vector_index_search_latency_usec                          |     latency      | Microseconds | Latency distribution (in microseconds) for flat vector index searches                                                                                                             |
+| hnsw_vector_index_search_latency_usec                          |     latency      | Microseconds | Latency distribution (in microseconds) for HNSW vector index searches                                                                                                             |
```
madolson: Why do only these get latency distributions and not full text?

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| search.ft-info-timeout-ms            | Number |   5000   | Timeout in milliseconds for FT.INFO command when CLUSTER or PRIMARY option is selected     |
+| search.ft-info-rpc-timeout-ms        | Number |   2500   | Timeout in milliseconds for RPC connections used by FT.INFO when CLUSTER OR PRIMARTY options is selected  |
+| search.local-fanout-queue-wait-threshold | Number |   50   | When this value is less than the average read queue wait time (in milliSeconds)  the local node is preferred in a fanout operation. |
+| search.thread-pool-wait-time-samples | Number |   100   | Sample queue size for thread pool wait time tracking   |
+| search.tag-min-prefix-length                  | Number  |   2   |               | Minimum number of characters required before trailing `*` in TAG wildcard queries (length excludes `*`) 
+| search.max-term-expansions           | Number |   200   | Maximum number of words to search in text operations (prefix, suffix, fuzzy) to limit memory usage |
+| search.search-result-buffer-multiplier | String |   1.5   | Multiplier for search result buffer size allocation    |
```
madolson: Why would this get modified? What is it used for?

topics/search-monitoring.md
```diff
@@ -0,0 +1,126 @@
...
+| search.drain-mutation-queue-on-save  | Boolean |   false   | Drain the mutation queue before RDB save.           |
+| search.query-string-depth            | Number |   1000   | Controls the depth of the query string parsing from the FT.SEARCH cmd        |
+| search.query-string-terms-count      | Number |   1000   | Controls the size of the query string parsing from the FT.SEARCH cmd (number of nodes in predicate tree)                 |
+| search.fuzzy-max-distance            | Number |   3   | Controls the maximum allowed edit distance for fuzzy search queries          |
+| search.max-vector-knn                | Number |   10000   | Controls the max KNN parameter for vector search       |
+| search.proximity-inorder-compat-mode | Boolean |   false   | Controls proximity iterator's inorder/overlap violation check logic (compatibility mode)    |
```
madolson: Why would this get modified? What is it used for?

madolson (changes requested):

commands/ft.aggregate.md
rlunar: I'll add the history to the [src/commands/ft.aggregate.json](https://github.com/valkey-io/valkey-search/blob/main/src/commands/ft.aggregate.json) similar to the [src/commands/set.json](https://github.com/valkey-io/valkey/blob/unstable/src/commands/set.json#L10) in the valkey core.

topics/search-monitoring.md
rlunar: Agree, this should be called **search-monitoring**. I would also steer away from _observability_ because it will give the wrong impression.

rlunar: After testing the module there's a few items missmatch, this is the summary:

=== MISSING FROM SERVER (documented but not in INFO output) ===
  MISSING: coordinator_bytes_in
  MISSING: coordinator_bytes_out
  MISSING: coordinator_client_get_global_metadata_failure_count
  MISSING: coordinator_client_get_global_metadata_failure_latency_usec
  MISSING: coordinator_client_get_global_metadata_success_count
  MISSING: coordinator_client_get_global_metadata_success_latency_usec
  MISSING: coordinator_client_search_index_partition_failure_count
  MISSING: coordinator_client_search_index_partition_failure_latency_usec
  MISSING: coordinator_client_search_index_partition_success_count
  MISSING: coordinator_client_search_index_partition_success_latency_usec
  MISSING: coordinator_server_get_global_metadata_failure_count
  MISSING: coordinator_server_get_global_metadata_failure_latency_usec
  MISSING: coordinator_server_get_global_metadata_success_count
  MISSING: coordinator_server_get_global_metadata_success_latency_usec
  MISSING: coordinator_server_listening_port
  MISSING: coordinator_server_search_index_partition_failure_count
  MISSING: coordinator_server_search_index_partition_failure_latency_usec
  MISSING: coordinator_server_search_index_partition_success_count
  MISSING: coordinator_server_search_index_partition_success_latency_usec
  MISSING: coordinator_threads_cpu_time_sec
  MISSING: flat_vector_index_search_latency_usec
  MISSING: hnsw_vector_index_search_latency_usec
  MISSING: prefiltering_requests_count

=== EXTRA IN SERVER (in INFO output but not documented) ===
  EXTRA: number_of_active_indexes
  EXTRA: number_of_active_indexes_indexing
  EXTRA: number_of_active_indexes_running_queries
  EXTRA: query_prefiltering_requests_cnt

For all the **Coordinator** metrics seems like we should split those into a separate table for cluster mode only; I wouldn't like to add a new column since we're already exceeding the available width on the table.

Image

rlunar (commented):
Provided feedback and a solution for the history on exisitng FT.* commands https://github.com/valkey-io/valkey-search/pull/909

topics/search-monitoring.md
KarthikSubbarao: @rlunar - There is a metrics update PR to remove unnecessary and incomplete metrics. The ones you found as missing from above are being actually removed

topics/search-telemetry.md
KarthikSubbarao: There is a metrics update PR to remove unnecessary and incomplete metrics. The ones mentioned above, are actually being removed. The Documentation PR was made knowing that these metrics are being removed

topics/search-monitoring.md
KarthikSubbarao: Yes, we are making it consistent by naming it as `prefiltering_requests_count` - this is a change which is about to be merged in, so the documentation PR was made based on that

topics/search-monitoring.md
KarthikSubbarao: Modules don't have a great visibility into the core config and their values. Unless they use ValkeyModule_Call and send the CONFIG command via a fake client, Modules do not know what the values are AFAIK.

I actually think this `search.hide-user-data-from-log` should be made a DEV config and removed from documentation. In the future, when Modules support better visibility, we will be able to check the actual config from the core and potentially rely on it

topics/search-monitoring.md
KarthikSubbarao: This was a new config merged in today. The bundle might be outdated as of now

topics/search-monitoring.md
KarthikSubbarao: > Agree, this should be called search-monitoring. I would also steer away from observability because it will give the wrong impression.

I was thinking of monitoring, but this also includes configs...

I am good with monitoring if there is no better name

topics/search-telemetry.md
KarthikSubbarao: Metric will be removed from here and from the source code. We will add it back when implemented

topics/search-monitoring.md
KarthikSubbarao: All of these metrics are "DEV" metrics. Valkey-search marks configs as DEV, HIDDEN, and APP.

DEV metrics cannot be changed (they are rejected) unless we are in debug-mode. This model was added here: https://github.com/valkey-io/valkey-search/pull/339/

topics/search-monitoring.md
KarthikSubbarao: It is the number of cores which are computed by this and assigned as the default value:
```
size_t GetPhysicalCPUCoresCount() {
  // Default
  size_t cpu_cores = std::thread::hardware_concurrency();
#if defined(__linux__) && defined(__aarch64__)
  auto res = helper::ExecuteCommand("/usr/bin/lscpu");
  if (!res.ok()) {
    VMSDK_LOG(WARNING, nullptr) << res.status().message()
                                << ". Returning value from "
                                   "std::thread::hardware_concurrency()";
  } else {
    const auto& lscpu_output = res.value();
    cpu_cores = helper::ParseLscpuOutput(lscpu_output);
  }
#elif defined(__linux__)
  std::ifstream cpuinfo("/proc/cpuinfo");
  if (!cpuinfo.is_open()) {
    VMSDK_LOG(NOTICE, nullptr)
        << "Could not read /proc/cpuinfo. Returning value from "
           "std::thread::hardware_concurrency()";
  } else {
    cpu_cores = helper::ParseCPUInfo(cpuinfo);
  }
#endif
  VMSDK_LOG(DEBUG, nullptr) << "Cores count is set to:" << cpu_cores;
  return cpu_cores;
}
```

topics/search-monitoring.md
KarthikSubbarao: Ack. I don't have the full context of all the metrics added in 1.0 since I was not involved in its  implementation/review/release until the end.

If there is a way to deprecate metrics in a major version, that could be an option to remove the metrics which we disagree with / don't find useful

topics/search-monitoring.md
KarthikSubbarao: I fixed the formatting of the tables

topics/search-monitoring.md
KarthikSubbarao: These were added in 1.0. The scope of the metric tracking here is limited to the vector search logic and core algorithms.

```
| search_flat_vector_index_search_latency_usec | Microseconds | Latency distribution for flat vector index searches |
| search_hnsw_vector_index_search_latency_usec | Microseconds | Latency distribution for HNSW vector index searches |
```

For Search commands (FT.SEARCH / FT.AGGREGATE for example) and their overall tracking, the command stats latency should be the source of truth for execution time and improvements in tracking regarding accuracies can be made there

topics/search-monitoring.md
KarthikSubbarao: Was this for `search_used_read_cpu` and `search_used_write_cpu`? If yes, I updated

topics/search-monitoring.md
KarthikSubbarao: This config provides a default option for delivering partial results when errors such as timeout occur. This applies a default behavior on commands which can be explicitly overriden by FT.INFO/FT.SEARCH commands using the SOMESHARDS/ALLSHARDS option

False in the config is ALLSHARDS. True in the config is SOMESHARDS

topics/search-monitoring.md
KarthikSubbarao: Useful. By not loading a hypothetical corrupted RDB, assuming the backfill (in case of RDB v1) or saved key list re-ingestion (v2) does not lead to a corruption, there would be no corruption in the index there after

topics/search-monitoring.md
madolson: I like that!

topics/search-monitoring.md
madolson: Put it in the doc

topics/search-monitoring.md
madolson: Update the doc, right now it's useless because it says nothing

topics/search-monitoring.md
madolson: The config name is misleading then. It's not that partial results are enabled, it's that they default to partial results.

topics/search-monitoring.md
madolson: Can we document that then. We don't just want to include unrelated information in the documentation

KarthikSubbarao: I can remove this `hnsw exception` section for now

topics/search-monitoring.md
KarthikSubbarao: Updated

rlunar:
I've taken a look at the new new version of **valkey-search** 

Commands are properly rendered using the **History** table.

2026-03-17_ft aggregate

New topic **search-monitoring** 

2026-03-17_search-monitoring

PR #425, opened 2026-03-20
fix(docs): correct LPOP example in lists documentation

gchape:
LPOP removes from the head, so the expected output after RPUSH bike:1 bike:2 bike:3 should return bike:1, bike:2, bike:3 in that order, not reverse.


zuiderkwast (approved):
Thanks!

Some years ago, some doc writers at Redis changed all the examples to be about bikes. I believe they made some mistakes. (Also, bikes are not always the best examples, but that's another problem. For sorted sets, the original was about highscore lists in gaming, which I believe are the more typical use case for sorted sets. Not bikes...)

PR #439, opened 2026-05-08
Add Database level ACL docs

dvkashapov:


commands/acl-log.md
```diff
@@ -3,6 +3,7 @@ The command shows a list of recent ACL security events:
 1. Failures to authenticate their connections with [`AUTH`](auth.md) or [`HELLO`](hello.md).
 2. Commands denied because against the current ACL rules.
 3. Commands denied because accessing keys not allowed in the current ACL rules.
+4. Commands denied because accessing a database not allowed in the current ACL rules.
```
zuiderkwast, line 9: We have yet another new ACL log event added in https://github.com/valkey-io/valkey/pull/1920.

It's not the same feature, but if you want to add it anyway, it'd be nice.

commands/acl-log.md
```diff
@@ -39,7 +40,7 @@ Entries are displayed starting from the most recent.
 Each log entry is composed of the following fields:
 
 1. `count`: The number of security events detected within a 60 second period that are represented by this entry.
-2. `reason`: The reason that the security events were logged. Either `command`, `key`, `channel`, or `auth`.
+2. `reason`: The reason that the security events were logged. Either `command`, `key`, `channel`, `database`, or `auth`. The `database` reason is reported when the user has no permission to access the database referenced by the command; the `object` field then contains the database id, or the command name when the command implicitly accesses every database (such as [`FLUSHALL`](flushall.md)).
```
zuiderkwast: The database reason is very detailed here but the other ones are not. We should follow the same style for all of them.

I don't mind that it explains it in detail, but we should do that also for the other reasons then.

The object should be described under the object bullet below instead.

We can also add the reason "tls-cert", which was introduced in the client-cert auth PR mentioned above.

commands/acl-setuser.md
```diff
@@ -72,6 +72,9 @@ This is a list of all the supported Valkey ACL rules:
 * `-<command>`: Remove the command to the list of commands the user can call. It can be used with `|` for blocking subcommands (e.g., "-config|set").
 * `-@<category>`: Like `+@<category>` but removes all the commands in the category instead of adding them.
 * `nocommands`: Alias for `-@all`. Removes all the commands, and the user is no longer able to execute anything.
+* `db=<id>[,<id>...]`: Adds the specified database id(s) to the list of databases the selector is allowed to access. Multiple ids are separated by commas, e.g. `db=0,1,2`. Adding any explicit `db=` rule clears the `alldbs` flag on the selector. See [database permissions](../topics/acl.md#database-permissions) for more information.
```
zuiderkwast: Adds or sets? If user previously had `db=0,1,2` and is modified with another ACL SETUSER with `db=3,4,5` will it add (result `0,1,2,3,4,5`) or overwrite (result `3,4,5`)?

If it's additive, it can be worth clarifying with an example below.

"Adding any explicit `db=` rule" is mostly confusing (I thought that you should use exactly `db=` without any databases here) so I suggest removing this part of the sentence.

```suggestion
* `db=<id>[,<id>...]`: Adds the specified database id(s) to the list of databases the selector is allowed to access.
  Multiple ids are separated by commas, e.g. `db=0,1,2`.
  Clears the `alldbs` flag on the selector.
  See [database permissions](../topics/acl.md#database-permissions) for more information.
```

commands/acl-setuser.md
```diff
@@ -94,3 +97,29 @@ OK
...
+OK
+```
+
+Different selectors can grant access on different databases:
+
+```
+127.0.0.1:6379> ACL SETUSER bob on nopass (db=0,1 +@write +select ~*) (db=2,3 +@read +select ~*)
+OK
+127.0.0.1:6379> ACL GETUSER bob
+...
```
zuiderkwast: Three dots is not very clear in an example. 😆 Please insert the actual reply of ACL GETUSER here. (Alternatively, don't include GETUSER at all in the example.)

commands/move.md
```diff
@@ -3,3 +3,8 @@ destination database.
 When `key` already exists in the destination database, or it does not exist in
 the source database, it does nothing.
 It is possible to use `MOVE` as a locking primitive because of this.
+
+The caller must have ACL permission to access the destination database. The
+source database is the currently selected one, so the user already had to be
+allowed to [`SELECT`](select.md) it. See
+[database permissions](../topics/acl.md#database-permissions).
```
zuiderkwast: > The source database is the currently selected one, so the user already had to be allowed to SELECT it.

Mentioning SELECT here is overly complicated. It's simpler to just say the caller must have access to the current and the destrination databases.

topics/acl.md
```diff
@@ -104,6 +104,12 @@ Allow and disallow Pub/Sub channels:
 * `allchannels`: Alias for `&*` that allows the user to access all Pub/Sub channels.
 * `resetchannels`: Flush the list of allowed channel patterns and disconnect the user's Pub/Sub clients if these are no longer able to access their respective channels and/or channel patterns.
 
+Allow and disallow logical databases:
+
+* `db=<id>[,<id>...]`: Add the specified database id(s) to the list of databases the user is allowed to access. Multiple ids are separated by commas, e.g. `db=0,1,2`. Adding any explicit `db=` rule clears the `alldbs` flag on the selector. See [database permissions](#database-permissions) for more information.
```
zuiderkwast: If this is additive, it's worth explaining that more explictly.

I read "any explicit `db=` rule clears the `alldbs` flag" as it's possible to specify `db=` without any databases. That's probably a no-op or a syntax error? The point is that `db=(any databases)` clears the alldbs flag so that the user only has access to the specified databases instead of all, which is kind of intuitive, but the text somehow complicates it. We can just skip the part "Adding any explicit `db=` rule" so that the sentence is just "Clears the `alldbs` flag on the selector."

Also, I recommend putting some linebreaks, so it's easier to review changes line by line now and in future changes. Git diffs are per line. Linebreaks in markdown don't render as linebreaks anyway (unless it's a double linebreak which means new paragraph).

```suggestion
* `db=<id>[,<id>...]`: Add the specified database id(s) to the list of databases the user is allowed to access.
  Multiple ids are separated by commas, e.g. `db=0,1,2`.
  Clears the `alldbs` flag on the selector.
  See [database permissions](#database-permissions) for more information.
```

topics/acl.md
```diff
@@ -437,6 +443,83 @@ For example, consider the following two commands:
...
+In addition to commands, keys, and Pub/Sub channels, ACL rules can restrict
+which databases (see [`SELECT`](../commands/select.md)) a user is
+allowed to access. By default a newly created selector has the `alldbs` flag
+set, meaning the user can use any database. Adding an explicit `db=` rule
+restricts the selector to only the listed databases.
+
+The default scope for database permissions is `alldbs`. The following rules
+control database access:
+
+* `db=<id>[,<id>...]`: Allow access only to the listed database ids.
```
zuiderkwast, line 462:
```suggestion
* `db=<id>[,<id>...]`: Allow access only to the listed database ids. (Clears the `alldbs` flag.)
```

topics/acl.md
```diff
@@ -437,6 +443,83 @@ For example, consider the following two commands:
...
+* `alldbs`: Allow access to every database (default for new selectors).
+* `resetdbs`: Remove every database from the allowed list and clear `alldbs`.
+
+For example, to create a user that may only operate on databases 0 and 1:
+
+```
+> ACL SETUSER alice on +@all ~* resetdbs db=0,1 nopass
```
zuiderkwast: `resetdbs` is not necessary when creating a new user. Is there any point to include it?

```suggestion
> ACL SETUSER alice on +@all ~* db=0,1 nopass
```

topics/acl.md
```diff
@@ -437,6 +443,83 @@ For example, consider the following two commands:
...
+1) "user alice on nopass sanitize-payload ~* resetchannels db=0,1 +@all"
+2) "user default on nopass ~* &* alldbs +@all"
+```
+
+Because adding `db=` automatically clears `alldbs`, `resetdbs` can be omitted:
+
+```
+> ACL SETUSER alice on +@all ~* db=0,1 nopass
+"user alice on nopass sanitize-payload ~* resetchannels db=0,1 +@all"
+```
```
zuiderkwast: I don't think this is necessary. It's very indirect. It's better to omit `resetdbs` in the first example above. If we want an example with `resetdbs` it should be about modifying an existing user to remove some database access.

topics/acl.md
```diff
@@ -437,6 +443,83 @@ For example, consider the following two commands:
...
+are checked specifically against the database list:
+
+Commands that explicitly reference a database id in their arguments require
+the user to have access to that id:
+
+* [`SELECT`](../commands/select.md) - target database id.
+* [`SWAPDB`](../commands/swapdb.md) - both database ids.
+* [`MOVE`](../commands/move.md) - destination database id (the source is the
+  currently selected database, which the user already had to be allowed to
+  `SELECT`).
```
zuiderkwast: If the source db is 0 and the user just connected, it can happen that it is not allowed to use db 0, right? We don't know that the user was allowed to SELECT it.

```suggestion
* [`MOVE`](../commands/move.md) - destination database id (and the source database which is the currently selected database).
```

madolson:
@dvkashapov Can you update the docs before GA?

topics/acl.md
dvkashapov: Applied.

topics/acl.md
dvkashapov: Added block about using `resetdbs` to remove database access from an existing user and deleted `resetdbs`.

topics/acl.md
dvkashapov: You're right that it's not needed, deleted here and added later block about using `resetdbs` to remove database access from an existing user.

topics/acl.md
dvkashapov: Done and added line about requiring at least one id 
```
* `db=<id>[,<id>...]`: Allow access only to the listed database ids.
  At least one id is required.
  This replaces any previously configured database ids and clears `alldbs`.
```

topics/acl.md
dvkashapov: It's not additive, cleared up wording around that. Added that at least one id is required because "db=" is syntax error. About clearing "alldbs" wrote this: "Clears the `alldbs` flag and any previously configured database ids on the selector." Applied line breaks

commands/move.md
dvkashapov: Right, done

commands/acl-setuser.md
dvkashapov: Added ACL LIST output here and explained a little bit about root selector to not confuse people about "alldbs -@all". This is a confusing topic, should we include explanation at all here? Or hope that user will understand without ACL LIST?

commands/acl-setuser.md
dvkashapov: You're right that "Adds" is not the right word. DB rules are not additive and are reset on every "db=", changed wording about that everywhere.

commands/acl-log.md
dvkashapov: Added "tls-cert" reason and explained a little bit about each object and reason.

commands/acl-log.md
dvkashapov: Added a little info about TLS clients auth and pubsub here, please check that I've not mistaken something for TLS event.

zuiderkwast: Pubsub was also missing? Good catch!

dvkashapov:
I think we're good to go with docs here

commands/acl-log.md
```diff
@@ -39,9 +42,21 @@ Entries are displayed starting from the most recent.
...
+   * `tls-cert`: automatic authentication from a TLS certificate failed because no matching user exists.
 3. `context`: The context that the security events were detected in. Either `toplevel`, `multi`, `lua`, `script`, or `module`.
-4. `object`: The resource that the user had insufficient permissions to access. `auth` when the reason is `auth`.
+4. `object`: The resource that the user had insufficient permissions to access. The value depends on `reason`:
+   * `command`: the command name.
+   * `key`: the key pattern.
+   * `channel`: the Pub/Sub channel.
+   * `database`: the database id, or the command name when the command implicitly accesses every database (such as [`FLUSHALL`](flushall.md)).
+   * `auth`: the literal string `AUTH`.
+   * `tls-cert`: the certificate that did not match any existing user.
```
zuiderkwast, line 59: Maybe I'm overthinking... The current docs is OK.

Below comment is just me thinking too much about this...

----

I think somehow we should say what the `object` is first in each line, instead of listing the `reason` first in each line here. The dependency on `reason` is maybe good to repeat everywhere to avoid confusion still. WDYT?

```suggestion
4. `object`: The resource that the user had insufficient permissions to access. The value depends on `reason`:
   * the command name, when `reason` is `command`
   * the key pattern, when `reason is `key`
   * the Pub/Sub channel, when `reason` is `channel`
   * the database id, or the command name when the command implicitly accesses every database (such as [`FLUSHALL`](flushall.md)), when `reason` is `database`
   * the literal string `AUTH` when `reason` is `auth`
   * the certificate that did not match any existing user, when `reason` is `tls-cert`.
```

A table would perhaps be even clearer, but that may be a too large change.

| reason | object |
|--|--|
| command | the command name |
| key | the key pattern |
| channel | the Pub/Sub channel |
| database | the database id, or the command name when the command implicitly accesses every database (such as [`FLUSHALL`](flushall.md)) |
| auth | the literal string `AUTH` |
| tls-cert | the certificate that did not match any existing user |

zuiderkwast (approved):

PR #441, opened 2026-05-19
Update Benchmark topic with latest valkey-benchmark.c changes

Andriciuc:
This PR focuses on updating the benchmark.md chapter from https://github.com/valkey-io/valkey-doc/issues/414 and using the latest updated valkey-benchmark.c file from https://github.com/valkey-io/valkey/pull/2823.

First commit changes:

add --duration and --warmup options
move --seed in correct option place
add two examples based on --duration, --warmup and --rps

Replaces #435

### DIFF
--- a/topics/benchmark.md
+++ b/topics/benchmark.md
@@ -60,6 +60,12 @@ In command arguments, the following placeholders are substituted:
 **`-n`** _requests_
 : Total number of requests (default 100000)
 
+**`--duration`** _seconds_
+: Run benchmark for specified number of seconds (mutually exclusive with -n)
+
+**`--warmup`** _seconds_
+: Run benchmark for specified warmup period before recording data.
+
 **`-d`** _size_
 : Data size of SET/GET value in bytes (default 3)
 
@@ -137,11 +143,11 @@ In command arguments, the following placeholders are substituted:
 **`-x`**
 : Read last argument from STDIN.
 
-**`--seed`** _num_
-: Set the seed for random number generator. Default seed is based on time.
-
 **`--rps`** _requests_
 : Limit the total number of requests per second. Default 0 (no limit).
+
+**`--seed`** _num_
+: Set the seed for random number generator. Default seed is based on time.
 
 **`--tls`**
 : Establish a secure TLS connection.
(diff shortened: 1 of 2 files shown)


Andriciuc:
The following changes have been done to the benchmarking chapter:

For Placeholders (Description section):

- __rand_int__: Added note that multiple occurrences within a single command each get different values
- Added __rand_1st__ through __rand_9th__ — same-value placeholders with independent counters
- Added __field:name__ — dataset field substitution, requires --dataset

New options added:

- --duration — time-based benchmark run, mutually exclusive with -n
- --warmup — pre-benchmark warmup period, stats discarded
- --sequential — sequential key replacement modifier for -r
- --rps — request rate limiter using token bucket algorithm
- --rfr — read-from-replica mode for cluster benchmarks
- --num-functions — controls function_load test (default 10)
- --num-keys-in-fcall — controls fcall test (default 1)
- --dataset — CSV/TSV file for field placeholder substitution
- --maxdocs — limits records loaded from dataset
- --fuzz — random command generation (testing only)
- --fuzz-mode — fuzzing modes: malformed-commands, config-commands
- --fuzz-loglevel — fuzzer log verbosity
- --rdma — RDMA connection support
- --mptcp — Multipath TCP support

Existing option corrections:

- -r: "expand" → "replace"; added per-occurrence behavior note
- --precision: default corrected from 0 to 3

Examples:

- Added time-based benchmark example (--warmup, --duration)
- Added rate-limited benchmark example (--rps)
- Added dataset benchmark example (--dataset, __field:name__)
- Removed \n\n"); C source artifact from transaction example

Keyspace output block:

- Removed backtick artifacts around <= in the sample output

topics/benchmark.md
```diff
@@ -21,10 +21,19 @@ In command arguments, the following placeholders are substituted:
...
+  Multiple occurrences within a single command each get different values.
+
+`__rand_1st__`
+: Similar to __rand_int__ but multiple occurrences will have the same
+  value. __rand_2nd__ through __rand_9th__ are also available.
 
 `__data__`
 : Replaced with data of the size specified by the -d option.
 
+`__field:name__`
```
VoletiRam: Can we be more specific here? The `name` here implies the column name from the csv/tsv file. Can we replace `name` with `colname`?

VoletiRam (approved):

topics/benchmark.md
Andriciuc: Fixed! Thank you!

PR #454, opened 2026-07-07
Add documentation for forkless operations

nitaicaro:
Add a documentation page covering forkless operations:

- How forkless operations work 
- Client blocking behavior during consistent and eventually-consistent iteration
- Configuration and triggering
- Inline replication for forkless full synchronization
- Monitoring via INFO fields
- Performance comparison with fork-based operations 

Forkless slot migration documentation will be added in the future once that work lands in the forkless branch (https://github.com/valkey-io/valkey/tree/forkless).

EDIT: Lots of changes introduced in the recent forkless PRs, will update this once we're all done and merged


madolson:
This is the right place, make it a little more high level since this file is intended for end users.

PR #469, opened 2026-08-17
Document the ACL DIGEST command

melancholictheory:
Documentation for `ACL DIGEST`, the command added in valkey-io/valkey#4446 and requested in valkey-io/valkey#4355.

`ACL DIGEST` returns a fingerprint of the ACL rules currently in effect, as a hexadecimal string. A controller that manages users through an ACL file can read it before and after an `ACL LOAD` and tell whether the file it wrote is the one now in effect. Comparing the rules instead does not work, because the server reports them in its own normalized form rather than the form they were written in.

This adds `commands/acl-digest.md` and the matching entries in `resp2_replies.json` and `resp3_replies.json`.

It should land after or together with the server PR. Until the command JSON exists in valkey, the Makefile leaves the page out of the build, since `commands` is the intersection of the Markdown pages and the JSON files under `VALKEY_ROOT/src/commands`, so nothing breaks in the meantime.

## Checks

The example output on the page was taken from a server built from the command branch rather than written by hand.

With `VALKEY_ROOT` pointed at that branch, `make` builds `_build/man/man3/acl-digest.3valkey.gz`. The summary, complexity, ACL categories, since version and reply description all resolve, and the links to `ACL LIST` and `ACL LOAD` render as man page references.

`utils/check-links.pl` reports no broken links, and spellchecker-cli is clean against `.spellcheckerrc.yml`.

### DIFF
--- a/commands/acl-digest.md
+++ b/commands/acl-digest.md
@@ -1,0 +1,30 @@
+Return a fingerprint of the ACL rules currently in effect, as a hexadecimal string.
+The digest depends only on the rules in effect and not on the order the users were
+created in, so two servers configured alike report the same digest.
+
+The digest covers everything [`ACL LIST`](acl-list.md) reports: the user names, their
+flags, their password hashes, and their command, key and channel rules.
+Any change to any user changes the digest.
+The reply is the same size however many users are configured.
+
+This is meant for confirming which revision of an ACL a server is running.
+A client that manages users through an ACL file can read the digest before and after an
+[`ACL LOAD`](acl-load.md) to tell whether the file it wrote is the one now in effect.
+Comparing the rules themselves is impractical, because the server reports them in its
+own normalized form rather than the form they were written in, so a client would need its
+own copy of the rule parser to compare them.
+
+## Examples
+
+```
+127.0.0.1:6379> ACL DIGEST
+"24ece0a4daea0bcb936b438ae6c9f36db76f57dfc70e00def7b2c26041c49822"
+127.0.0.1:6379> ACL SETUSER alice on >alice ~key:* +get
+OK
+127.0.0.1:6379> ACL DIGEST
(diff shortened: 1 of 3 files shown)


zuiderkwast (approved):
LGTM, thanks!

Waiting for https://github.com/valkey-io/valkey/pull/4446 to be merged (and for the major decision to be approved) before merging this.

melancholictheory:
The command merged as valkey-io/valkey@e23ccb31 on 31 August, so this is unblocked.

I re-read the page against what actually shipped and the text holds.

Order independence is by construction: the implementation computes SHA256 over each user's name and rule string and XORs the per-user digests together, so no ordering enters the result.

"Covers everything `ACL LIST` reports" is accurate for the same reason. The input per user is the name plus `ACLDescribeUser`, which is the string `ACL LIST` prints, so flags, password hashes and the command, key and channel rules are all inside the digest.

The fixed-size claim holds too, since the reply is a SHA256 in hex regardless of how many users are configured.

The `ACL DELUSER` example is worth keeping. Returning to the original digest after the user is removed is the XOR round trip, and it shows the property without needing to explain it.

One thing the page does not state, deliberately, is the version. The command shipped with `since` 9.2.0 in `acl-digest.json`, and the site renders that from the command metadata rather than from the page.

commands/acl-digest.md
```diff
@@ -0,0 +1,30 @@
+Return a fingerprint of the ACL rules currently in effect, as a hexadecimal string.
+The digest depends only on the rules in effect and not on the order the users were
+created in, so two servers configured alike report the same digest.
+
+The digest covers everything [`ACL LIST`](acl-list.md) reports: the user names, their
+flags, their password hashes, and their command, key and channel rules.
+Any change to any user changes the digest.
```
zuiderkwast, line 7: We don't need to mention that Roles will affect the digest too, even though it will, assuming that feature will land in 9.2 too.

PR #471, opened 2026-08-19
Update sorted set docs: skiplist replaced by B+ tree

### LINKED ISSUE #466: Clean up skiplist references after the sorted set B+ tree rewrite
The sorted set skiplist was replaced with a B+ tree in valkey-io/valkey#4359 ("Replace the ZSET skiplist with a B+ tree behind an ordered index"). Two places in valkey-doc still describe the old structure.

### 1. `commands/object-encoding.md:32`

```
    - `skiplist`, normal sorted set encoding.
```

`OBJ_ENCODING_SKIPLIST` became `OBJ_ENCODING_BTREE` (same value, 7), and `object.c` now returns the literal string `btree`, so `OBJECT ENCODING` on a large sorted set answers `btree`.

Since the change is only in `unstable` and has not shipped in a release yet, both names should be listed, following the existing `zipmap` precedent in the same file:

```
    - `btree`, normal sorted set encoding.
    - `skiplist`, the name used for the normal sorted set encoding before it was replaced by a B+ tree.
    - `listpack`, a space-efficient encoding used for small sorted sets.
```

### 2. `topics/sorted-sets.md:51`

```
Implementation note: Sorted sets are implemented via a
dual-ported data structure containing both a skip list and a hash table, so
every time we add an element Valkey performs an O(log(N)) operation.
```

The skip list is gone. `zset` now holds a hashtable plus an `OrderedIndex` backed by a B+ tree (`fbtree.c`). Only the structure name needs to change. The O(log(N)) claim is still correct.

### Note

`btree` is not in `wordlist` (only `btree1-az` is, from a cluster example). The spellcheck workflow runs over `commands/*.md` and `topics/*.md`, so the word needs a wordlist entry or CI will fail.

### Checked and not affected

* `zset-max-listpack-entries` / `zset-max-listpack-value` and the listpack encoding for small sorted sets are untouched, so `topics/memory-optimization.md` and `commands/config-get.md` are correct as written.
* All O(log(n)) and O(log(n)+m) complexity statements still hold.
* The only other user-visible change in the commit is a `DEBUG OBJECT` log line ("Skiplist level" to "Index height"), which valkey-doc does not document.

Separately, `commands/scan.md:90` says "Hashes and Sorted Sets encoded as ziplists". That is stale from the listpack conversion rather than this rewrite, so it is out of scope here.

---

*This was generated by AI but verified, with love, by a human.*

rainsupreme:
The sorted set skiplist was replaced with a B+ tree in valkey-io/valkey#4359. Update two places in valkey-doc:

1. commands/object-encoding.md: OBJECT ENCODING now returns 'btree' for normal sorted sets. Keep 'skiplist' as a historical note following the existing zipmap precedent.

2. topics/sorted-sets.md: Replace 'skip list' with 'B+ tree' in the implementation note. The O(log N) claim remains correct.

3. wordlist: Add 'btree' so the spellcheck workflow passes.

Closes valkey-io/valkey-doc#466

### DIFF
--- a/commands/object-encoding.md
+++ b/commands/object-encoding.md
@@ -29,7 +29,8 @@ Valkey objects can be encoded in different ways:
 
 * Sorted Sets can be encoded as:
 
-    - `skiplist`, normal sorted set encoding.
+    - `btree`, normal sorted set encoding.
+    - `skiplist`, the name used for the normal sorted set encoding before it was replaced by a B+ tree.
     - `listpack`, a space-efficient encoding used for small sorted sets.
 
 * Streams can be encoded as:
--- a/topics/sorted-sets.md
+++ b/topics/sorted-sets.md
@@ -48,7 +48,7 @@ With sorted sets it is trivial to return a list of hackers sorted by their
 birth year because actually *they are already sorted*.
 
 Implementation note: Sorted sets are implemented via a
-dual-ported data structure containing both a skip list and a hash table, so
+dual-ported data structure containing both a B+ tree and a hash table, so
 every time we add an element Valkey performs an O(log(N)) operation. That's
 good, but when we ask for sorted elements Valkey does not have to do any work at
 all, it's already sorted. Note that the `ZRANGE` order is low to high, while the `ZREVRANGE` order is high to low:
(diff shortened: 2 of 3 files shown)


madolson (approved):
🐝 tree!

PR #479, opened 2026-09-01
Add CI guard and Makefile targets for module API reference docs

murphyjacob4:
The `topics/modules-api-ref.md` page is programmatically generated from doc comments in `src/module.c` in the Valkey core repository using `utils/generate-module-api-doc.rb`. Manual edits made directly to `topics/modules-api-ref.md` are easily lost during regeneration or lead to discrepancies between the core codebase comments and the documentation.

This commit adds:

1. A GitHub Actions workflow (`.github/workflows/check-modules-api-ref.yml`) that verifies `topics/modules-api-ref.md` matches the generated output from `src/module.c` in `valkey-io/valkey` (at the release tag or unstable) whenever `topics/modules-api-ref.md` is modified in a PR.
2. Handles skipping the revoked 8.0.8 tag (which was mistakenly tagged on unstable) so unreleased APIs are not erroneously attributed to 8.0.8.
3. Makefile targets `generate-modules-api-ref` and `verify-modules-api-ref` to easily regenerate and verify the documentation locally.
4. Documentation in `README.md` explaining that `topics/modules-api-ref.md` is auto-generated and instructions on how to contribute module doc updates via `src/module.c`.

Note that we should still only do the bump on release, or the CI will have issues on unrelated PRs when the `unstable` branch incorporates new module API changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,3 +268,23 @@ distclean:
 install: install-man
 
 uninstall: uninstall-man
+
+# ---- Modules API Reference ----
+
+.PHONY: generate-modules-api-ref verify-modules-api-ref
+
+generate-modules-api-ref:
+	@git -C $(VALKEY_ROOT) tag -d 8.0.8 2>/dev/null || true
+	ruby $(VALKEY_ROOT)/utils/generate-module-api-doc.rb > topics/modules-api-ref.md
+
+verify-modules-api-ref:
+	@git -C $(VALKEY_ROOT) tag -d 8.0.8 2>/dev/null || true
+	@ruby $(VALKEY_ROOT)/utils/generate-module-api-doc.rb > topics/modules-api-ref.md.tmp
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -268,3 +268,23 @@ distclean:
 install: install-man
 
 uninstall: uninstall-man
+
+# ---- Modules API Reference ----
+
+.PHONY: generate-modules-api-ref verify-modules-api-ref
```
madolson, line 274: The docs today are primarily generated here https://github.com/valkey-io/valkey-release-automation/blob/4c1744dd7daf0e64879e4f19776d4c468a3ef8e2/.github/workflows/update-valkey-doc.yml#L126-L129. I'm fine adding another step, but would be nice to update the other location to use this makefile if we now also have it here.

madolson (approved):

Makefile
murphyjacob4: Makes sense. I can update it to use this new process once we submit

